PR #17, opened 2022-06-14
[BackPort MineTick] Fix Occasional Client Side Unloading of Chunk 0,0

mechoriet:
This helped me fix a ghost/void chunk on 0,0 cords.

When some one walk outside of the view distance and then back to 0,0.
Or sometimes it also already happens when they join its already void

There are probably more cases this could occur but these where the most reported before we ported this to our inhouse fork


uRyanxD:
A small note: You need to change the patch number (from 0063 to 0064)

mechoriet:
@hpfxd little tip to keep the repo cleaner use `Squash and Merge` next time,
So it them will only create one commit with all the changes
with the name of the Pull Request

then you would only to make sure that Pull Request names are good and reflect the change
like paper and you would have less noise commits that don't actually do something 

and double commits for the same thing like with the `Cache chunk coordinations` Pull Request

hpfxd:
@mechoriet I typically prefer creating a merge commit as opposed to squashing to preserve history, but I suppose it doesn't matter that much for this project, especially since most changes are patch files anyway. I'll squash-merge PRs in the future.

PR #40, opened 2022-11-08
[WIP] Port "Use fastutil Collections" patch

uRyanxD:
https://github.com/TacoSpigot/TacoSpigot/blob/version/1.8.8/PaperSpigot-Server-Patches/0021-Use-Fastutil-Collections.patch

### DIFF
--- a/patches/server/0068-Use-Fastutil-Collections.patch
+++ b/patches/server/0068-Use-Fastutil-Collections.patch
@@ -1,0 +1,200 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Aikar <aikar@aikar.co>
+Date: Sun, 6 Nov 2022 14:36:47 -0300
+Subject: [PATCH] Use Fastutil Collections
+
+Swap out CraftBukkit LongObjectHashMap with Long2ObjectOpenHashMap
+Swap out Integer key HashMap for a Int2ObjectOpenHashMap For ChunkProviderServer
+
+Also swap out datawatcher collections for their fastutil equivelents.
+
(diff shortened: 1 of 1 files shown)


hpfxd:
This patch increases the final JAR size from 26MiB to 43MiB due to the inclusion of fastutil. I've considered including fastutil in the past, but I'm not sure if the optimizations gained from using it would be worth the extra size.

uRyanxD:
I agree with you, I'll leave this PR as Draft, if anyone has a test that proves it's worth it, we'll discuss implementing it

mechoriet:
"Amaranth, the creator of LongObjectHashMap, tested it vs fastutil and determined fastutil to be 3x faster
and could not create anything faster than fastutil.

These collections are super fast as seen 
https://web.archive.org/web/20170701185749/https://java-performance.info/hashmap-overview-jdk-fastutil-goldman-sachs-hppc-koloboke-trove-january-2015/"

But I believe there so blazing fast cause they use more static(fastutil) dedicated lists for that purpose then the more generic hashmaps and lists that java offers less weird lookups and type optimized lookups.

if that makes any sense

uRyanxD:
I found 1 aikar project that apparently only contains minimal fastutil classes, maybe we could use it?
https://github.com/aikar/fastutil-lite

mechoriet:
If we include it it should be the whole thing imo.
cause else you could get confusion that classes don't exist while expected it is also for some people using the full spigot jar in there dev env for example

also found some other places where a these maps/lists with better initial sizes could be useful mainly in NBTTagCompound
which is kinda used everywhere

PR #47, opened 2022-11-21
Multiple memory and performance patches

mechoriet:
Instead of flooding the pullrequestlist for every little patch
I found it probbly better since these shouldn't break anything it has been running for a few days and had moving bots walking around + me flying around without it ever deadlocking

- minor Furnace patch - if bucket in slot is empty then return earliest possible
- avoid recreating block break animation packet class - since they are always the same but looped thru players inside a distance from action happening
- don't update entity tracker if no players are even there to process - lowers idle cpu usage
- add getTypeIfLoaded - getType only if chunk is loaded
- optimised Light Comparison 
- Optimize getChunkAt calls - could sometimes be called for the same chunk multiple times so a small cache of the key so we quickly find it again inside the list 
- add Velocity Compression of packets and decompression
- Use less resources for collisions - by removing the use of BlockPosition classes
- unused creating of BlockPosition objects
- Get rid of some repeated isChunkLoaded checks - by just calling areNeighborsLoaded since we already know the data 
- dont check collisions if movement is very-very minimal 
- MovementCache
- optimize GameRules TreeMap to LinkedHashMap - saw it high in spark profiles 

also got more patches but those involve fastutil maps which could reduce memory usage even more


patches/server/0077-Velocity-Compression-from-paper.patch
```diff
@@ -0,0 +1,449 @@
...
++import com.velocitypowered.natives.compression.VelocityCompressor;
+ import io.netty.buffer.ByteBuf;
+ import io.netty.channel.ChannelHandlerContext;
+ import io.netty.handler.codec.MessageToByteEncoder;
+@@ -7,16 +8,28 @@ import java.util.zip.Deflater;
+ 
+ public class PacketCompressor extends MessageToByteEncoder<ByteBuf> {
+ 
+-    private final byte[] a = new byte[8192];
++    private final byte[] a; // Paper
```
uRyanxD, line 155: incorrect comment, it must be // PandaSpigot

patches/server/0077-Velocity-Compression-from-paper.patch
```diff
@@ -0,0 +1,449 @@
...
+     private final Deflater b;
+     private int c;
++    private final com.velocitypowered.natives.compression.VelocityCompressor compressor; // Paper
```
uRyanxD, line 158: other mistake

patches/server/0077-Velocity-Compression-from-paper.patch
```diff
@@ -0,0 +1,449 @@
...
++import com.velocitypowered.natives.compression.VelocityCompressor;
++import com.velocitypowered.natives.util.MoreByteBufUtils;
+ import io.netty.buffer.ByteBuf;
+ import io.netty.buffer.Unpooled;
+ import io.netty.channel.ChannelHandlerContext;
+@@ -13,9 +15,14 @@ public class PacketDecompressor extends ByteToMessageDecoder {
+     private final Inflater a;
+     private int b;
+ 
++    private final com.velocitypowered.natives.compression.VelocityCompressor compressor; // Paper
```
uRyanxD, line 267: other xd

patches/server/0077-Velocity-Compression-from-paper.patch
```diff
@@ -0,0 +1,449 @@
...
+@@ -8,17 +10,30 @@ import javax.crypto.Cipher;
+ 
+ public class PacketDecrypter extends MessageToMessageDecoder<ByteBuf> {
+ 
+-    private final PacketEncryptionHandler a;
++    private final VelocityCipher cipher; // PandaSpigot
+ 
+-    public PacketDecrypter(Cipher cipher) {
+-        this.a = new PacketEncryptionHandler(cipher);
++    public PacketDecrypter( VelocityCipher cipher) { // Paper
```
uRyanxD, line 345: eeeeh

patches/server/0077-Velocity-Compression-from-paper.patch
```diff
@@ -0,0 +1,449 @@
...
+ import io.netty.handler.codec.MessageToByteEncoder;
+-import javax.crypto.Cipher;
++import io.netty.handler.codec.MessageToMessageEncoder;
+ 
+-public class PacketEncrypter extends MessageToByteEncoder<ByteBuf> {
++import javax.crypto.Cipher;
+ 
+-    private final PacketEncryptionHandler a;
++public class PacketEncrypter extends MessageToMessageEncoder<ByteBuf> {
++    private final VelocityCipher cipher; // Paper
```
uRyanxD, line 394: more

patches/server/0077-Velocity-Compression-from-paper.patch
andreasdc: hater alert

mechoriet:
> If I remember correctly MovementCache patch was useless and the conditions were impossible to achieve, because players had always "falling down" movement. I may be wrong, had to be debugged of course.

true after checking with debug messages if its like a farm with entities not moving it uses the cache but 95% of the time it not so it doesn't feel worth it

andreasdc:
> > If I remember correctly MovementCache patch was useless and the conditions were impossible to achieve, because players had always "falling down" movement. I may be wrong, had to be debugged of course.
> 
> true after checking with debug messages if its like a farm with entities not moving it uses the cache but 95% of the time it not so it doesn't feel worth it

Btw did you check if for mobs it's the same situation? The "falling" movement blocking the cache?

uRyanxD:
I think it's not a good idea to add the patch "0081-Optimize-TileEntity-Ticking.patch", because, it wasn't finalized by Aikar, and two diff's (TileEntityEnderChest, TIleEntityChest) are missing from the patch you ported

mechoriet:
> Btw did you check if for mobs it's the same situation? The "falling" movement blocking the cache?

Haven't had much time to look at this again, if x/y/z had changed by what amount

mechoriet:
> I think it's not a good idea to add the patch "0081-Optimize-TileEntity-Ticking.patch", because, it wasn't finalized by Aikar, and two diff's (TileEntityEnderChest, TIleEntityChest) are missing from the patch you ported

those 2 diff don't look to be needed anymore since they already got moved in this commit f1faf9ca from paper
they moved the sound things from ontick check to the actual startOpen and closeContainer methods 
for both of the mentioned classes therefor those files didn't got changed

hpfxd:
Thank you for the PR. I've cleaned up and manually merged most of these in c7b9962ec75f099f117a32ecd7a44eff93cf1175...e1f79e4d476be8df50349e9d7ed0af6454f3b303. I've left out a few for now that I don't believe will provide too much of a benefit; or that are quite large/need changes and should probably be in their own PR. In general, multiple PRs over one large one is typically preferred as it's easier to organize and review, even if it floods the list.

mechoriet:
> 

Thought this was better else it would've been a big flood but if that's more preferred sure
cause I could this takes ages to get all checked and modified if it is/was big like this

uRyanxD:
> I think it's not a good idea to add the patch "0081-Optimize-TileEntity-Ticking.patch", because, it wasn't finalized by Aikar, and two diff's (TileEntityEnderChest, TIleEntityChest) are missing from the patch you ported

I know that this PR has been closed for a long time, however, apparently I made a mistake, the patch was finalized by aikar apparently, in the past I had only found the beta version of it, now I found a finalized one, I just don't know why the patch was not updated to 1.8
https://github.com/PaperMC/Paper-1.7/blob/master/Spigot-Server-Patches/0031-Optimize-TileEntity-Ticking.patch

PR #55, opened 2022-12-04
Addition to Prevent loading chunks patch

mechoriet:


mechoriet:
Could also add this to the existing 0075-Avoid-loading-chunks-in-multiple-places if that's more desired to at least somewhat limit how much patches we get overtime

hpfxd:
that would be preferred, yes :)

PR #56, opened 2022-12-04
saving empty scoreboards / adding non-player's to scoreboards with config option

mechoriet:
Could prevent messes created by not very smart plugins not cleaning up after them self


uRyanxD:
LGTM, and if you can include this patch in this PR that would be nice (https://github.com/TacoSpigot/TacoSpigot/blob/version/1.8.8/PaperSpigot-Server-Patches/0013-Disable-Scoreboards-for-non-players-by-default.patch)

mechoriet:
> LGTM, and if you can include this patch in this PR that would be nice (https://github.com/TacoSpigot/TacoSpigot/blob/version/1.8.8/PaperSpigot-Server-Patches/0013-Disable-Scoreboards-for-non-players-by-default.patch)

added that patch since it touches same area/ same intention

uRyanxD (approved):
LGTM, need rebase

PR #58, opened 2022-12-04
Reduce Chunk Hashing

mechoriet:
If the fastutil gets added and implemented for `unloadQueue` this patch could be reordered after implementation of `FU` and updated to also include those extra hashing's since they all for the chunkkey

but this already reduces it on heavy chunk loading and unloading servers since its twice less sometimes even triple less hash calculations for the same X and Z cords


uRyanxD:
LGTM

PR #59, opened 2022-12-04
Optimize Light Recalculations

mechoriet:
If chunk is already known why not pass it and only for unknown places look it up
so it avoids a search for a known chunk
should help a bit in high chunk count servers
since it has to go thru a massive list possible multiple times


iamnoksio:
NEED TO BE ADDED!!!

uRyanxD (approved):

PR #62, opened 2022-12-05
Optimized nether portal search

mechoriet:
this patch starts searching more limited since the search isn't returned when there is a option found to possibly find a more optimal one,
which would and could result in a lot of wasted chunk loads/generation cause nether is not always prerendered compared to overworlds 

if we haven't found one yet since its possibly above a lake so we do the default search radius (default Mechanics)

-edit reworded-


uRyanxD:
Is the chance that the method finds a portal at a small distance great? If not, it would just delay the method of finding a portal

mechoriet:
with how it currently works if keeps searching the whole 128 search radius even if there is a result earlier,
which should in 99% of cases be in the near vicinity of original location (except for possible lava lakes) so it would exit/return result back faster

so its a fast vs slow return
and by the time it scans for the slower/longer version those chunks it loaded in the fast/shorter run should still be loaded so they should be fast then

maybe the only thing that could happen it has to reload those chunks by gc timing of that build in chunk-gc system
could be improved later if implemented by having a offset function added but idk if that is worth it since you could miss a bit in between the search ranges of 9 to 11 if not done right

uRyanxD (approved):
I rebased the PR and did a cleanup, LGTM

PR #67, opened 2022-12-11
TCP Options

uRyanxD:
Note: TCPFastOpen does not improve performance much, in some tests done by a friend with bot attacks according to him it even got worse, if you agree I think it's a good thing not to add

### DIFF
--- a/patches/server/0085-TCP-Options.patch
+++ b/patches/server/0085-TCP-Options.patch
@@ -1,0 +1,44 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Sculas <sculas@github.com>
+Date: Sun, 11 Dec 2022 14:32:50 -0300
+Subject: [PATCH] TCP Options
+
+
+diff --git a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotConfig.java b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotConfig.java
+index ae3cdaca275b1a50c5c37c3dd0021c4bd579d373..c69ef08217fb8cc9e31bc6ca3a2e5c71f3836094 100644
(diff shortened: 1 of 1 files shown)


mechoriet:
From what I just read it could improve in servers that are backed by a proxy on the initial login or server switches since it doesn't need to reestablish a new connection fully since a sorta cookie gets set on both ends that they are "friends"/good folk
https://en.wikipedia.org/wiki/TCP_Fast_Open 

so it could start sending chunk data for example sooner then without it
since the buildup of a tcp connection is drastically shortened

but for standalone servers it not very useful since both sides need to have this enabled to work, and every player is in a different location and a highly likely chance there client doesn't have this enabled at all 

so this is more a behind a proxy kind of optimization for faster initial contact

xIsm4:
> Note: TCPFastOpen does not improve performance much, in some tests done by a friend with bot attacks according to him it even got worse, if you agree I think it's a good thing not to add

I don't think it's right to use TFO in the backend, it would be a good pull-request in BungeeCord (because Velocity already has it) and compress chunks faster (making the connection faster) which is the point of this. As for the bot attack, I don't understand why PandaSpigot should mitigate it, it's not made for that.

mechoriet:
I meant it in the proxy software(bungee/velocity) <-> lobbies/game modes kind of situations.
then if proxy software has this also enabled then subsequent connections to a lobby and gamemode should see improvement
the other side clients <-> pandaspigot won't see any benefit from this since its mainly for sequent connections that provide the token that is already known.
So it doesn't need to do the full handshake to establish the connection.

If I'm understanding the little reading I did about it?

xIsm4:
> 

That's why i think is worth pull-request on BungeeCord. Velocity already had it

uRyanxD:
I'm closing this PR because I don't have time to work on it

PR #68, opened 2022-12-12
Kick on receiving invalid creative action packet

uRyanxD:
Ported from https://github.com/ProjectKig/KigPaper/blob/ver/1.8.8/Spigot-Server-Patches/0170-Kick-after-receiving-invalid-creative-action-packet.patch

### DIFF
--- a/patches/server/0085-Kick-after-receiving-invalid-creative-action-packet.patch
+++ b/patches/server/0085-Kick-after-receiving-invalid-creative-action-packet.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: RoccoDev <hey@rocco.dev>
+Date: Mon, 12 Dec 2022 15:30:23 -0300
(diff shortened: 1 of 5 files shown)


uRyanxD:
Rebased

uRyanxD:
Rebased

uRyanxD:
Rebased

uRyanxD:
Rebased

uRyanxD:
Rebased

uRyanxD:
Rebased

anzz1:
This would probably be better as a configuration option, similar to antiXray.
As in if there is some cheat clients out there which this is trying to fight against(?)
For some server that never uses creative mode could safely set the option on, but there is the [HumanEntity.setGameMode()](https://hub.spigotmc.org/javadocs/bukkit/org/bukkit/entity/HumanEntity.html#setGameMode(org.bukkit.GameMode)) command and I reckon that if some plugin changed a player's gamemode, due to network delay the client still thinking it is in creative mode while server thinks its already switched out of it, would end up kicking him.

Issue #70, opened 2022-12-19
[SUGGESTION] impl net/techcable/tacospigot/event/entity/SpawnerPreSpawnEvent


Paladinow:
i need this event for my stack mobs plugin

PR #71, opened 2022-12-19
Add SpawnerPreSpawnEvent

mechoriet:
This solved #70 
only one addition to the tacospigot one is that
I've added a check to see if there are listeners to avoid firing if no one is listening to it


patches/server/0085-Implement-SpawnerPreSpawnEvent.patch
```diff
@@ -0,0 +1,39 @@
...
+ import org.bukkit.craftbukkit.event.CraftEventFactory;
+ import org.bukkit.event.entity.CreatureSpawnEvent;
+ import org.bukkit.event.entity.SpawnerSpawnEvent;
+@@ -91,6 +92,19 @@ public abstract class MobSpawnerAbstract {
+                 boolean flag = false;
+ 
+                 for (int i = 0; i < this.spawnCount; ++i) {
++                    // PandaSpigot start - add check to avoid firing if non listeners
++                    if(SpawnerPreSpawnEvent.getHandlerList().getRegisteredListeners().length > 0) {
++                        // TacoSpigot start
```
uRyanxD, line 26: Taco 😆

Paladinow: bruh

mechoriet: @uRyanxD cause its a 1.8.8 patch from taco https://github.com/TacoSpigot/TacoSpigot/blob/version/1.8.8/PaperSpigot-Server-Patches/0022-Implement-SpawnerPreSpawnEvent.patch

and added only the start and end to the change I made to limit broadcasting that event if there are listeners for it 
since it cancelable and spawning code is generally hotter since it ran very often so kind of  a waste to construct a class every time it tries to spawn things

PR #75, opened 2022-12-27
reduce NBTTagCompound footprint

mechoriet:
since no inclusion of fastutil this is the maximum of reduction that is possible


iamnoksio:
good one

iamnoksio:
ADDD IT

patches/server/0085-Reduce-memory-footprint-of.patch
```diff
@@ -0,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: mechoriet <kevinworm92@gmail.com>
+Date: Tue, 27 Dec 2022 14:20:44 +0100
+Subject: [PATCH] Reduce memory footprint of  NBTTagCompound
```
uRyanxD, line 4: double space in name

PR #82, opened 2023-01-13
Update Github Actions

uRyanxD:
Let it compile on multiple versions of java
 - Probably shouldn't improve performance, however it is useful for Panda development
 
 
 - I marked this PR as WIP because I forgot to edit the check.yml, I edit it later


mechoriet:
Don't know about this one since then PandaSpigot will have 3 different jars floating around and probbly a load of questions why there jar isn't working since they using the wrong version

and don't really think this adds much then longer build time and it doesn't change anything apart from requiring a higher java version to be started
cause the jvm already optimizes the code by in-lining and other things on compile of the class on runtime
to my knowledge.

uRyanxD:
^^ I agree, I edited the wrong file I was going to add it just to compile in PRs (check.yml) to make sure all used methods exist in new javas

PR #91, opened 2023-02-04
Only call the event if there is a listener that uses it

uRyanxD:


iamnoksio:
this is useless, even if the event is fire it still do nothing

iamnoksio:
this is useless, even if the event is fire it still do nothing

uRyanxD:
> this is useless, even if the event is fire it still do nothing

Yes, even if it is called, it doesn't do anything, but it doesn't make sense to call it if it doesn't do anything

iamnoksio:
> > this is useless, even if the event is fire it still do nothing
> 
> Yes, even if it is called, it doesn't do anything, but it doesn't make sense to call it if it doesn't do anything

even if it doesnt make sens to you, it still dont change any performances/fix something, and if it was absolutely right you'd do it for every single event

PR #105, opened 2023-03-14
Do not call the events if they are not listened to by a plugin

SaithTime:
I see many spigot's fork adding options to disable some events (like the PlayerMoveEvent) to get better performance.

However, if calling an event is so expensive, why call them if no plugins listen to them?

To test I added the event names to a list when they are registered and I check when they are called if they are in the list.
I know it's a bit dirty, I don't really have time to test in depth the impact on performance and the problems it can cause, but I wanted to share this optimization track anyway


iamnoksio:
If an event if fired it still does nothing if no plugin are using the event, close this pull request

PR #110, opened 2023-04-15
Fix Entity Tracker Rounding

ThatApplePieGuy:
the vanilla entity tracker always floors when converting coordinates to an int before sending them to the client, this patch rounds properly so player positions are slightly more accurate clientside. more info here: https://www.youtube.com/watch?v=Ba2TCTf8_nU

### DIFF
--- a/patches/server/0086-Fix-Entity-Tracker-Rounding.patch
+++ b/patches/server/0086-Fix-Entity-Tracker-Rounding.patch
@@ -1,0 +1,61 @@
+From: necrozma <necrozma999@gmail.com>
+Date: Sat, 8 Apr 2023 04:32:10 -600
+Subject: [PATCH] Fix Entity Tracker Rounding
+
+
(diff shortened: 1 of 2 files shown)


hpfxd (approved):

andreasdc:
Was it compared to the client's position with old and new version of rounding coordinates?

ThatApplePieGuy:
here's a crude visualization, each set of points is the player's position on the block after walking into that corner. without this patch players are effectively misplaced towards 0,0, so some players may have extra reach on others depending on the direction they're facing.
ingame

andreasdc:
It's still not the actual position right?

ThatApplePieGuy:
no it's not the actual position, some precision is always lost as a limitation of how minecraft sends the coordinates

andreasdc:
Not good, is there a way to fix that? So basically it is stil not precise but not precise in the same direction, right?

ThatApplePieGuy:
there's not any solution to this, the packet itself is limited to sending coordinates in multiples of 0.03125

andreasdc:
Maybe it's good to report it to mojang? But probably they won't fix it anyway :D

PR #112, opened 2023-04-15
Use default vanilla knockback settings

uRyanxD:
mechoriet forgot a part of the config


mechoriet (approved):
LGTM tried to find all the default values must have missed one

PR #126, opened 2023-04-28
Fix chest checking

uRyanxD:
I got this fix from PR 74, but the other PR has other changes, and I need more time to analyze them, so I separated this change in this PR to be implemented faster


mechoriet:
yeah that's fine
patch looked a bit messy now to check without having to dig/search the places

mechoriet (approved):

PR #129, opened 2023-04-29
Add support for io_uring

skbeh:
This only works when not using unix domain socket
and compression is disabled because
the netty io_uring transport does not support
unix domain socket and the client can not decode
compressed packets sent with io_uring.


MachineBreaker:
Disable this by default

skbeh:
@MachineBreaker Done.

uRyanxD:
Hello, thanks for the pullrequest, however, we decided not to add support for IO_URING for now as it is an experimental feature and for security reasons as Kenny said [here](https://github.com/PaperMC/Paper/pull/9141#issuecomment-1913517266).

andreasdc:
What's not secure with this solution? I heard that io_uring brings great performance.

PR #142, opened 2023-05-18
Make falling block height nerf respect doEntityDrops

uRyanxD:


hpfxd:
i'd prefer if these two patches were kept separate, instead of merging them into one. they don't seem to be related (other than being fixes to Paper patches). other than that, lgtm

uRyanxD:
> i'd prefer if these two patches were kept separate, instead of merging them into one. they don't seem to be related (other than being fixes to Paper patches). other than that, lgtm

I separated the two patches now

hpfxd (approved):

PR #143, opened 2023-05-20
Add log4j-slf4j2-impl

cheaterSK:
I noticed that when using plugins using log4j (mainly SlimeWorldManager), they can't find the slf4j2 implementation and thus switch to NO-OP. This patch fixes this by shading the implementation artifact as it should've been done like that in the first place. I have also updated log4j to the latest versions.


uRyanxD:
I made some changes and rolled back the log4j update, because the console was spamming "WARN StatusConsoleListener Using package checking to find plugins is deprecated and will be removed in a future release" and I couldn't find anything about the cause or how to resolve it that, LGTM

uRyanxD (approved):

PR #146, opened 2023-06-18
Fix banning players by name

P3ridot:
So without this missing method, server uses default `JsonList#a` method implementation which returns just toString for key (in this case it's GameProfile). By adding this method, we return result of `GameProfileBanList#b` which looks like that 
```java
protected String b(GameProfile gameprofile) {
    return gameprofile.getId().toString();
}
```
It's done in the same way in `WhiteList` and there everything works


HanielCota:
That is good! like

mechoriet:
Seems like this method just got deleted instead of decompile fixed https://github.com/hpfxd/PandaSpigot/blob/master/patches/server/0001-Fix-Decompilation-errors.patch#L621-L632

uRyanxD:
Yeah

PR #147, opened 2023-07-01
multiple patches + some cleanup around touched patches

mechoriet:
- port asm code generation for event listeners instead of reflection(should improve performance)

- moved place to check if there are listeners to the SimplePluginManager  just before fired

- port Performance & Concurrency Improvements to Permissions no double lookup

- packet limiter fix compile warning

- avoid double io lookup if file exists 

- don't allow pickup delay to go below 0

- Check channel before reading and change isOpen calls to
 IsActive

remove some more places where blockpositions are used often and can just be 3 int's x,y,z


mechoriet (commented):
this is the same state as before only removed the patch if no listener skip but some plugins do funky stuff

patches/server/0098-Avoid-double-I-O-operation-on-load-player-file.patch
```diff
@@ -0,0 +1,29 @@
...
+diff --git a/src/main/java/net/minecraft/server/WorldNBTStorage.java b/src/main/java/net/minecraft/server/WorldNBTStorage.java
+index ba13f3f20bb86f667178f51f4d3d1df63f5acf2d..7218f0c0dda6bdbacfcd0c5dd00796fdc27155ec 100644
+--- a/src/main/java/net/minecraft/server/WorldNBTStorage.java
++++ b/src/main/java/net/minecraft/server/WorldNBTStorage.java
+@@ -209,7 +209,8 @@ public class WorldNBTStorage implements IDataManager, IPlayerFileData {
+             File file = new File(this.playerDir, entityhuman.getUniqueID().toString() + ".dat");
+             // Spigot Start
+             boolean usingWrongFile = false;
+-            if ( org.bukkit.Bukkit.getOnlineMode() && !file.exists() ) // PaperSpigot - Check online mode first
++            boolean normalFile = file.exists() && file.isFile(); // PandaSpigot
```
MachineBreaker, line 16: File#isFile already checks if the file exists, if you don't remove the exists() call then you're still doing double I/O operation

patches/server/0100-Check-channel-before-reading-and-change-isOpen-calls.patch
```diff
@@ -0,0 +1,48 @@
...
+-        if (this.channel != null && !this.channel.isOpen()) {
++        if (this.channel != null && !this.channel.isActive()) { // PandaSpigot
+             if (!this.p) {
+                 this.p = true;
+                 if (this.j() != null) {
+@@ -565,7 +565,9 @@ public class NetworkManager extends SimpleChannelInboundHandler<Packet> {
+     }
+ 
+     protected void channelRead0(ChannelHandlerContext channelhandlercontext, Packet object) throws Exception { // CraftBukkit - fix decompile error
++        if(g()) { // PandaSpigot start - Check if channel is opened before reading packet
```
MachineBreaker, line 43: This is not required as #a method already checks if the channel is open.

patches/api/0027-Performance-Concurrency-Improvements-to-Permissions-.patch
uRyanxD, line 1: public PermissionAttachment addAttachment(Plugin plugin, int ticks)
public PermissionAttachment addAttachment(Plugin plugin, String name, boolean value, int ticks)

These methods were made synced in the original patch (https://github.com/PaperMC/Paper/blob/ver/1.12.2/Spigot-API-Patches/0134-Performance-Concurrency-Improvements-to-Permissions.patch) any reason why they aren't included?

patches/server/0100-Check-channel-before-reading-and-change-isOpen-calls.patch
mechoriet: but if I remove the check it will 100% throw npe's it the channel is already dead

MachineBreaker: Yeah, resolved

patches/server/0101-reduce-new-blockposition-generation.patch
```diff
@@ -0,0 +1,72 @@
...
++                return CraftMagicNumbers.getBlock(previous.getTypeId()).fromLegacyData(previous.getRawData());
++            }
++        }
++        return null;
++    }
++    // PandaSpigot end
++
+     // PandaSpigot start - Add getTypeIfLoaded
+     public IBlockData getTypeIfLoaded(BlockPosition blockposition) {
++        int x = blockposition.getX(); // PandsSpigot - temp set xyz till blockpos nuke
```
uRyanxD, line 43: Pands

patches/server/0101-reduce-new-blockposition-generation.patch
```diff
@@ -0,0 +1,72 @@
...
++        int y = blockposition.getY(); // PandsSpigot - temp set xyz till blockpos nuke
```
uRyanxD, line 44: Pands

patches/server/0101-reduce-new-blockposition-generation.patch
```diff
@@ -0,0 +1,72 @@
...
++        int z = blockposition.getZ(); // PandsSpigot - temp set xyz till blockpos nuke
```
uRyanxD, line 45: Pands

uRyanxD (changes requested):

patches/server/0085-Implement-SpawnerPreSpawnEvent.patch
```diff
@@ -7,22 +7,17 @@ Added registered listeners check to avoid alloc if no plugins are
...
+index e54f7edbd1d834391f6760ea7b82cf6af4ddf994..b1780579735e025f469f86642bf303a9ccb9b207 100644
 --- a/src/main/java/net/minecraft/server/MobSpawnerAbstract.java
 +++ b/src/main/java/net/minecraft/server/MobSpawnerAbstract.java
-@@ -91,6 +91,17 @@ public abstract class MobSpawnerAbstract {
+@@ -91,6 +91,12 @@ public abstract class MobSpawnerAbstract {
                  boolean flag = false;
  
                  for (int i = 0; i < this.spawnCount; ++i) {
 +                    // PandaSpigot start - SpawnerPreSpawnEvent
-+                    if (net.techcable.tacospigot.event.entity.SpawnerPreSpawnEvent.getHandlerList().getRegisteredListeners().length > 0) {
```
uRyanxD, line 18: As you removed the patch that skipped events that had no registered listeners, you can re-add for this event

uRyanxD: If my mind is not failing you added the check back in however removed the use of "patches/api/0025-Add-a-call-helper-to-Event.patch" in the patch

patches/server/0073-Add-World-Util-Methods.patch
uRyanxD, line 1: This patch requires a cleanup (Add missing comments, fix comments, etc).

I also noticed that other than the alternate isLoaded() method and the IChunkProvider change, the rest doesn't see to have been used for anything, removing them seems like a good idea

patches/server/0005-Backport-PlayerHandshakeEvent-from-Paper.patch
```diff
@@ -20,8 +20,7 @@ index 215648ece51b9e0316bae17de60abccfd40f8b35..b7615b1472470805e6ee2a49dd8d4cdc
 +                if (com.destroystokyo.paper.event.player.PlayerHandshakeEvent.getHandlerList().getRegisteredListeners().length != 0) { // Hello? Can you hear me?
 +                    java.net.SocketAddress socketAddress = this.b.l;
 +                    String hostnameOfRemote = socketAddress instanceof java.net.InetSocketAddress ? ((java.net.InetSocketAddress) socketAddress).getHostString() : InetAddress.getLoopbackAddress().getHostAddress();
-+                    com.destroystokyo.paper.event.player.PlayerHandshakeEvent event = new com.destroystokyo.paper.event.player.PlayerHandshakeEvent(packethandshakinginsetprotocol.hostname, hostnameOfRemote, !proxyLogicEnabled);
-+                    org.bukkit.Bukkit.getPluginManager().callEvent(event);
++                    com.destroystokyo.paper.event.player.PlayerHandshakeEvent event = org.bukkit.craftbukkit.event.CraftEventFactory.callEvent(new com.destroystokyo.paper.event.player.PlayerHandshakeEvent(packethandshakinginsetprotocol.hostname, hostnameOfRemote, !proxyLogicEnabled));
```
uRyanxD, line 23: Any reason to use CraftEventFactory instead of "patches/api/0025-Add-a-call-helper-to-Event.patch" like you did in SpawnerPreSpawnEvent?

patches/server/0022-EntityMoveEvent.patch
```diff
@@ -18,8 +18,7 @@ index 72c7e6fc8bb0a71877d6759af44d39030bcf51f5..cd95aa17291d31e3cebe9f6488a43923
 +                org.bukkit.Location from = new org.bukkit.Location(this.world.getWorld(), this.lastX, this.lastY, this.lastZ, this.lastYaw, this.lastPitch);
 +                org.bukkit.Location to = new org.bukkit.Location (this.world.getWorld(), this.locX, this.locY, this.locZ, this.yaw, this.pitch);
 +                io.papermc.paper.event.entity.EntityMoveEvent event = new io.papermc.paper.event.entity.EntityMoveEvent((LivingEntity) this.getBukkitEntity(), from, to.clone());
-+                Bukkit.getPluginManager().callEvent(event);
-+                if (event.isCancelled()) {
++                if (CraftEventFactory.callEvent(event).isCancelled()) {
```
uRyanxD, line 21: same question for here too

patches/server/0037-Sound-events.patch
```diff
@@ -5,17 +5,16 @@ Subject: [PATCH] Sound events
...
-@@ -944,20 +944,57 @@ public abstract class World implements IBlockAccess {
+@@ -944,20 +944,56 @@ public abstract class World implements IBlockAccess {
      }
  
      public void makeSound(Entity entity, String s, float f, float f1) {
 +        // PandaSpigot start - EntitySoundEvent
-+        com.hpfxd.pandaspigot.event.sound.EntitySoundEvent event = new com.hpfxd.pandaspigot.event.sound.EntitySoundEvent(
-+                entity.bukkitEntity, s, f, f1);
-+        this.getServer().getPluginManager().callEvent(event);
++        com.hpfxd.pandaspigot.event.sound.EntitySoundEvent event = CraftEventFactory.callEvent(new com.hpfxd.pandaspigot.event.sound.EntitySoundEvent(
```
uRyanxD, line 16: same question for here too

patches/server/0073-Add-World-Util-Methods.patch
```diff
@@ -45,7 +65,129 @@ index b75d78da71f24a0b02bf709d026389add0556489..56d9a764bbf521f512e959efaa864c44
...
+@@ -297,7 +303,12 @@ public abstract class World implements IBlockAccess {
+         return this.getType(blockposition).getBlock().getMaterial() == Material.AIR;
+     }
+ 
++    // PandaSpigot start - xyz version of isLoaded
++    public boolean isLoaded(int x, int y, int z) {
++        return getChunkIfLoaded(x >> 4, z >> 4) != null;
++    }
+     public boolean isLoaded(BlockPosition blockposition) {
++    // PandaSpigot stop
```
uRyanxD, line 77: One of the things I wanted to say to fix in my previous review was this, "stop" instead of "end" and also need to move the comment 1 line up

patches/server/0073-Add-World-Util-Methods.patch
```diff
@@ -45,7 +65,129 @@ index b75d78da71f24a0b02bf709d026389add0556489..56d9a764bbf521f512e959efaa864c44
...
+         return getChunkIfLoaded(blockposition.getX() >> 4, blockposition.getZ() >> 4) != null; // PandaSpigot
+     }
+ 
+@@ -362,25 +373,92 @@ public abstract class World implements IBlockAccess {
+         return this.chunkProvider.getOrCreateChunk(i, j);
+     }
+ 
+-    public boolean setTypeAndData(BlockPosition blockposition, IBlockData iblockdata, int i) {
++
++    // PandaSpigot start - add captured versions
```
uRyanxD, line 87: It was particularly about these methods that I said about removing, because, I didn't see any use (Maybe there is any but I've already marked many patches as seen)

uRyanxD (changes requested):

patches/server/0005-Backport-PlayerHandshakeEvent-from-Paper.patch
```diff
@@ -21,8 +21,7 @@ index 215648ece51b9e0316bae17de60abccfd40f8b35..b7615b1472470805e6ee2a49dd8d4cdc
 +                    java.net.SocketAddress socketAddress = this.b.l;
 +                    String hostnameOfRemote = socketAddress instanceof java.net.InetSocketAddress ? ((java.net.InetSocketAddress) socketAddress).getHostString() : InetAddress.getLoopbackAddress().getHostAddress();
 +                    com.destroystokyo.paper.event.player.PlayerHandshakeEvent event = new com.destroystokyo.paper.event.player.PlayerHandshakeEvent(packethandshakinginsetprotocol.hostname, hostnameOfRemote, !proxyLogicEnabled);
-+                    org.bukkit.Bukkit.getPluginManager().callEvent(event);
-+                    if (!event.isCancelled()) {
++                    if (event.callEvent()) {
```
uRyanxD, line 24: TODO: Was this check reversed this whole time, or am I going crazy?

Coccocoahelper:
Bump

Issue #128, opened 2023-04-29
Build fails on latest Arch Linux


skbeh:
```console
Applying: Add player view distance API
Applying: Add BeaconEffectEvent
Applying: Add PlayerInitialSpawnEvent
Applying: Timings v2
Applying: Add Location support to tab completers (vanilla feature missing in CraftBukkit)
Applying: Made EntityDismountEvent Cancellable
Applying: Automatically disable plugins that fail to load
Applying: Expose server CommandMap
Applying: Graduate bungeecord chat API from spigot subclasses
Applying: Player Tab List and Title APIs
Applying: Add exception reporting event
Applying: Fix ServerListPingEvent flagging as Async
Applying: Fix ServerListPingEvent flagging as Async
Using index info to reconstruct a base tree...
M       src/main/java/org/bukkit/event/server/ServerEvent.java
M       src/main/java/org/bukkit/event/server/ServerListPingEvent.java
Falling back to patching base and 3-way merge...
No changes -- Patch already applied.
Patches applied cleanly to PaperSpigot-API
Cloning into 'PaperSpigot-Server'...
done.
Resetting PaperSpigot-Server to Spigot-Server...
HEAD is now at c90d379ea Don't sleep between chunk saves
Applying patches to PaperSpigot-Server...
Applying: POM Changes
Applying: PaperSpigot config files
Applying: mc-dev imports
Applying: Vanished players don't have rights
Applying: Allow undead horse types to be leashed
Applying: Configurable squid spawn ranges
Applying: Configurable damage multiplier for PvP blocking
Applying: Configurable cactus and reed natural growth heights
Applying: Configurable baby zombie movement speed
Applying: Configurable fishing time ranges
Applying: Allow nerfed mobs to jump
Applying: Toggle for player interact limiter
Applying: Player Exhaustion Multipliers
Applying: Add configurable despawn distances for living entities
Applying: Allow for toggling of spawn chunks
Applying: Drop falling block and tnt entities at the specified height
Applying: Configurable speed for water flowing over lava
Applying: Player affects spawning API
Applying: Show 'PaperSpigot' in client crashes, server lists, and Mojang stats
Applying: Metrics
Applying: Remove invalid mob spawner tile entities
Applying: Optimize TileEntity Ticking
Applying: Remove certain entities that fly through unloaded chunks
Applying: Configurable strength and weakness effect modifiers
Applying: Further improve server tick loop
Applying: Only refresh abilities if needed
[...]

PR #149, opened 2023-07-03
Fix build error with Arch Linux

skbeh:
In Arch Linux, the `cp` command throws some error messages when using `-n` and ignoring files, potentially because it is newer.

Ignore the error for now.

Close https://github.com/hpfxd/PandaSpigot/issues/128.


uRyanxD:
Does this still happen? @skbeh

PR #167, opened 2023-10-06
EntityPickupItemEvent 

mechoriet:


patches/server/0099-Implement-EntityPickupItemEvent.patch
```diff
@@ -0,0 +1,68 @@
...
+                 }
++                // PandaSpigot start - EntityPickupItemEvent
++                // Call newer event afterwards
++                org.bukkit.event.entity.EntityPickupItemEvent entityEvent = new org.bukkit.event.entity.EntityPickupItemEvent(entityhuman.getBukkitEntity(), (org.bukkit.entity.Item) this.getBukkitEntity(), remaining);
++                // event.setCancelled(!entityhuman.canPickUpLoot); TODO
++                this.world.getServer().getPluginManager().callEvent(entityEvent);
++                if (entityEvent.isCancelled()) {
++                    return;
++                }
++                itemstack.count = canHold + remaining;
```
hpfxd, line 64: Is this right? Above, `itemstack.count` is set to `canHold` before calling PlayerPickupItemEvent, but is changed to `canHold + remaining` before EntityPickupItemEvent is called, and set again to this value afterwards.

hpfxd:
I've cleaned the patch up a bit and implemented the event in places where it was missing (picking up arrows and insentient mobs). I've not had the time to test this in-game. If anyone has tested and can confirm this is working as it should, please share results :)

patches/server/0099-Implement-EntityPickupItemEvent.patch
mechoriet: changed to only run that after both methods it was confusing cause I was looking at 2 diff sources

uRyanxD:
I tested the event using myself (player entity) and it worked perfectly, however, I couldn't test it with mobs (Mobs picking up items depends on luck and I wasn't lucky)

uRyanxD (approved):
I tested the event with a Zombie and myself (player entity) and it worked perfectly for me.

If anyone else is going to test the event, here is a command to generate a mob that takes the item (without needing luck):
/summon Zombie ~ ~1 ~ {CanPickUpLoot:1b}

Thanks ApplePies for sending me the command.

patches/server/0099-Implement-EntityPickupItemEvent.patch
```diff
@@ -0,0 +1,74 @@
...
+                 if (event.isCancelled()) {
+                     return;
+                 }
++                // PandaSpigot start - EntityPickupItemEvent
++                // Call newer event afterwards
++                org.bukkit.event.entity.EntityPickupItemEvent entityEvent = new org.bukkit.event.entity.EntityPickupItemEvent(entityhuman.getBukkitEntity(), event.getItem(), 0);
++                // event.setCancelled(!entityhuman.canPickUpLoot); TODO
++                this.world.getServer().getPluginManager().callEvent(entityEvent);
++
++                if (entityEvent.isCancelled()) {
```
uRyanxD, line 21: We can reduce the diff here using the method added in commit 6af956b

patches/server/0099-Implement-EntityPickupItemEvent.patch
```diff
@@ -0,0 +1,74 @@
...
+             }
+ 
+-            if (flag && this.a(itemstack)) {
++            // PandaSpigot start - EntityPickupItemEvent
++            boolean canPickup = flag && this.a(itemstack);
++            org.bukkit.event.entity.EntityPickupItemEvent entityEvent = new org.bukkit.event.entity.EntityPickupItemEvent((org.bukkit.entity.LivingEntity) this.getBukkitEntity(), (org.bukkit.entity.Item) entityitem.getBukkitEntity(), 0);
++            entityEvent.setCancelled(!canPickup);
++            this.world.getServer().getPluginManager().callEvent(entityEvent);
++
++            if (!entityEvent.isCancelled()) {
```
uRyanxD, line 43: same thing here

patches/server/0099-Implement-EntityPickupItemEvent.patch
```diff
@@ -0,0 +1,74 @@
...
+ 
+                 if (event.isCancelled()) {
+                     return;
+                 }
++                // PandaSpigot start - EntityPickupItemEvent
++                // Call newer event afterwards
++                org.bukkit.event.entity.EntityPickupItemEvent entityEvent = new org.bukkit.event.entity.EntityPickupItemEvent(entityhuman.getBukkitEntity(), (org.bukkit.entity.Item) this.getBukkitEntity(), remaining);
++                // event.setCancelled(!entityhuman.canPickUpLoot); TODO
++                this.world.getServer().getPluginManager().callEvent(entityEvent);
++                if (entityEvent.isCancelled()) {
```
uRyanxD, line 67: same thing here

hpfxd (approved):

patches/server/0099-Implement-EntityPickupItemEvent.patch
mechoriet: can't here since it needs altering with setcanceled

PR #168, opened 2023-10-07
WIP: FastUtil

mechoriet:
got them now split up for visibility per class for patch

so it's a tiny bit more clear what was does then we can merge some in the end to condence them a bit


mechoriet:
3acd639 build 
[paperclip.zip](https://github.com/hpfxd/PandaSpigot/files/12838876/paperclip.zip)

HeathLoganCampbell:
From my experience there was no real major improve from this change

Issue #178, opened 2024-01-16
ServerConnect error


gitcarlx:
[ServerConnect] Task #16 for ServerConnect v0.1 generated an exception
java.lang.NoSuchMethodError: 'org.bukkit.entity.Player[] org.bukkit.Bukkit.getOnlinePlayers()'
        at com.gmail.emiel.topixx.ServerConnect$1.run(ServerConnect.java:50) ~[?:?]
        at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftTask.run(CraftTask.java:59) ~[patched_1.8.8
pigot-116]
        at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftScheduler.mainThreadHeartbeat(CraftSchedule
atched_1.8.8.jar:git-PandaSpigot-116]
        at net.minecraft.server.v1_8_R3.MinecraftServer.B(MinecraftServer.java:890) ~[patched_1.8.8.
igot-116]
        at net.minecraft.server.v1_8_R3.DedicatedServer.B(DedicatedServer.java:412) ~[patched_1.8.8.
igot-116]
        at net.minecraft.server.v1_8_R3.MinecraftServer.A(MinecraftServer.java:815) ~[patched_1.8.8.
igot-116]
        at net.minecraft.server.v1_8_R3.MinecraftServer.run(MinecraftServer.java:648) ~[patched_1.8.
Spigot-116]
        at net.minecraft.server.v1_8_R3.MinecraftServer.lambda$spin$0(MinecraftServer.java:133) ~[pa
:git-PandaSpigot-116]
        at java.lang.Thread.run(Thread.java:842) [?:?]

uRyanxD:
See [this](https://github.com/hpfxd/PandaSpigot/issues/174#issuecomment-1984559629)

Issue #174, opened 2023-12-23
java.lang.NoSuchMethodError: org.bukkit.Server.getOnlinePlayers()


victortedesco:
[15:39:28 WARN]: org.apache.commons.lang.UnhandledException: Plugin ChestShop v3.8.12 generated an exception while executing task 122
        at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:57)
        at org.github.paperspigot.ServerSchedulerReportingWrapper.run(ServerSchedulerReportingWrapper.java:23)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
        at java.lang.Thread.run(Thread.java:750)
Caused by: java.lang.NoSuchMethodError: org.bukkit.Server.getOnlinePlayers()[Lorg/bukkit/entity/Player;
        at com.Acrobot.ChestShop.Metrics.postPlugin(Metrics.java:335)
        at com.Acrobot.ChestShop.Metrics.access$400(Metrics.java:57)
        at com.Acrobot.ChestShop.Metrics$1.run(Metrics.java:223)
        at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftTask.run(CraftTask.java:59)
        at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:53)
        ... 4 more
        
This only happens with PandaSpigot, on vanilla Paper everything works fine.

andreasdc:
> ChestShop

It shouldn't break the plugin, you can try to disable metrics, I don't see this version on github.

victortedesco:
https://github.com/hpfxd/PandaSpigot/blob/c105f01a66997ac770dc2c59a18852950938d66c/patches/server/0049-Metrics.patch#L130C4-L144C4

Wasn't this made to solve this problem?

liorsl:
The patch you sent seems to modify the built in metrics class, but your plugin seems to be using its own which doesn't properly support 1.8.

uRyanxD:
[...]

Issue #159, opened 2023-08-25
error multiverse


f3f5-dev:
caused by pandaspigot. tested with flamepaper

https://pastes.dev/dizwwjyIgC

uRyanxD:
This error is due to the removal of legacy mappings (More information in issue #7).
You can use this version of MultiverseCore (https://dev.bukkit.org/projects/multiverse-core/files/2428161) to avoid this error, it was made for 1.9 but works perfectly in 1.8

Issue #152, opened 2023-07-12
Server crash


Ancienty:
Hello,

I don't know if I'm supposed to submit this as an error, I just switched to Panda and I don't really know if this is caused by PandaSpigot.
However the output on the console said I should report it, so here I am.

Sorry in advance if I shouldn't have submitted this issue.

```
[03:49:41] [Craft Scheduler Thread - 441/WARN]: Exception in thread "Craft Scheduler Thread - 441" 
[03:49:41] [Craft Scheduler Thread - 441/WARN]: org.apache.commons.lang.UnhandledException: Plugin CombatLog v1.8.11-hack generated an exception while executing task 75
	at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:57)
	at org.github.paperspigot.ServerSchedulerReportingWrapper.run(ServerSchedulerReportingWrapper.java:23)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:823)
Caused by: java.lang.NoSuchMethodError: org/bukkit/Server.getOnlinePlayers()[Lorg/bukkit/entity/Player; (loaded from file:/root/Skyblock/cache/patched_1.8.8.jar by sun.misc.Launcher$AppClassLoader@9a664ecd) called from class com.jackproehl.plugins.Metrics (loaded from file:/root/Skyblock/plugins/CombatLog-1.8.11-fix.jar by org.bukkit.plugin.java.PluginClassLoader@bd2b491c).
	at com.jackproehl.plugins.Metrics.postPlugin(Metrics.java:192)
	at com.jackproehl.plugins.Metrics.access$4(Metrics.java:186)
	at com.jackproehl.plugins.Metrics$1.run(Metrics.java:119)
	at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftTask.run(CraftTask.java:59)
	at org.bukkit.craftbukkit.v1_8_R3.scheduler.CraftAsyncTask.run(CraftAsyncTask.java:53)
	... 4 more

[03:49:43] [Async Chat Thread - #23/INFO]: 6.16K ┃ YETKILI vanks47 ► her gün kravat takıyoruz bavulla okula gidiyorduk
[...]

PR #188, opened 2024-03-08
Remap final jar using Spigot deprecation mappings

### LINKED ISSUE #7: Conflict with Multiverse Core (Works normally with Paper)
Replacing server.jar from PaperSpigot to Panda causes NoSuchMethodException, full server logs:
https://paste.gg/p/anonymous/117dfa7c512f4696aea9ede5af713387

- Using latest 37 build

hpfxd:
This patch re-introduces the remapping of `_INVALID_` deprecated methods performed by upstream, as well as transforming old `int` health-related setters to have the synthetic modifier (plugins should be using the `double` methods).
This is done to provide binary compatibility with plugins built against very old versions of the Bukkit API.

This implementation for PandaSpigot uses a modified version of the [`RemapTask` in `gradle-specialsource`](https://github.com/typst-io/gradle-specialsource/blob/8b038c46d1b9e8a6a523ad7278c4b07b078650dd/src/main/kotlin/io/typecraft/gradlesource/RemapTask.kt), licensed under Apache 2.0.

Fixes #7, #152, #159, #174, #178

### DIFF
--- a/build-logic/build.gradle.kts
+++ b/build-logic/build.gradle.kts
@@ -9,4 +9,5 @@ repositories {
 
 dependencies {
     implementation("io.sigpipe:jbsdiff:1.0")
+    implementation("net.md-5:SpecialSource:1.11.3")
 }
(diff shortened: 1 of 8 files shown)


uRyanxD (approved):
I tested compiling the project and changes using MultiverseCore, it works fine for me

PR #214, opened 2024-07-18
[WIP] Update to Netty 4.2.x

uRyanxD:


iamnoksio:
To really update Netty y'all need to remove net.minecraft.server from every class to only use netty package, otherwise it's useless, y'all gonna still use Netty from Minecraft server dependencies

Issue #231, opened 2024-08-17
the ender dragon bug


MichaelKozlovOfficial:
After I summon an Ender Dragon, this error is thrown: https://pastebin.com/aTtGnrB7

edit by @uRyanxD: When you spawn him in water

rejuvenate7:
doesnt have to be when spawned, the exception is thrown whenever any ender dragon touches water

PR #235, opened 2024-08-22
fix: EntitySoundEvent null pointer exception

rejuvenate7:
fixes issue #231


MichaelKozlovOfficial:
deleted the patch, fixed the issues? Don't be ridiculous.

MichaelKozlovOfficial:
why don't you fix patch instead of delete?

iamnoksio:
> why don't you fix patch instead of delete?

why backporting instead of actually fixing it? skill issue?

iamnoksio:
https://stackoverflow.com/questions/218384/what-is-a-nullpointerexception-and-how-do-i-fix-it if you ever wonder what a NPE is

iamnoksio:
it's only because EnderDragon sounds is returning a null entity; just set the EnderDragon as entity --'

uRyanxD:
> it's only because EnderDragon sounds is returning a null entity; just set the EnderDragon as entity --'
> 
> NAH F THE EDIT BRO

I think the problem is that the event is using entity.bukkitEntity instead of the entity.getBukkitEntity() method

uRyanxD:
@rejuvenate7,
Thanks for the pull request, however this is not the solution. The solution is to use getBukkitEntity() instead of entity.bukkitEntity, I will send a pullrequest with the fix later

PR #240, opened 2024-09-01
Fix console messages being black by default

### LINKED ISSUE #204: PandaSpigot terminal is broken/weird/incorrect
Examples:
Earning an achievement
Marking a player as an operator

Paper:
https://prnt.sc/YUlf4yjYrhb6

Panda:
https://prnt.sc/NLVOj73PDpXt

Probable cause:
https://github.com/hpfxd/PandaSpigot/blob/master/patches/server/0040-Use-TerminalConsoleAppender-for-console-improvements.patch

uRyanxD:
Fix #204

I wasted a lot of time trying to find the problem, which was so obvious, initially I tried to put WHITE here like in Paper, but I put the ) in the wrong place and that's why it didn't work lol


Rabbit0w0:
This fixed nothing but changed black to white. You should set it to RESET

uRyanxD:
> This fixed nothing but changed black to white. You should set it to RESET

You're right, however, I'm out of time to change this, PR's are welcome

PR #274, opened 2025-01-03
Remove random damage for protection.

sprytex:
This is with a diamond sword (Sharpness 4) and no enchantments on the armor. (Critical hits are off in the core.) (ME VS REJOMY) 
 
image

Without armor, we take the same amount of damage. If you wear armor with the Protection enchantment, the damage becomes random.

2025-01-03_19 06 04
![image
](https://github.com/user-attachments/assets/cd06543e-d4e3-4ecb-a7ec-784c741bf1ac)


sprytex:
@regomy made a fix

iamnoksio:
Is it a vanilla behavior?

iamnoksio:
Don't forget to mention me ^^

iamnoksio (commented):
3 times the same code at different places isnt looking good

PR #284, opened 2025-02-03
Accurate pearl Y level tracking 

ptbnate:
This feature is highly requested by most PvP based servers. In vanilla mc, when a player throws a pearl at an opponent, they teleport to the pearl's landing position, often leaving them in the air. This feature ensures the player teleports to the opponent’s Y level instead.


hpfxd:
It seems like this PR contains two unrelated patches. It would be helpful if each patch were in its own PR to merge and review separately, on their own branches.

ptbnate:
Gotcha.

PR #288, opened 2025-02-09
Fix incorrect head height for non-players

ptbnate:


ptbnate:
I'll merge this into the old one.

PR #318, opened 2025-05-20
Fix chest display bug and reset NaN entity values

mikheilkozlovi:


uRyanxD (changes requested):

uRyanxD:
Apparently you edited the patch manually, don't do that, it breaks the patch (as seen here), use our script ./panda edit server or api

mikheilkozlovi:
Not really interested, i have my own fork

PR #321, opened 2025-05-24
Fix chest visual open bug

### LINKED ISSUE #312: Visual open chest issue when you're making a lot of CPS while clicking
Can you add a patch to prevent the visual open chest issue that occurs when you're making a lot of CPS while clicking?

Image

https://github.com/ruViolence/Reaper/blob/ver/1.12.2/patches/server/0138-feat-drop-use-packets-while-inside-the-inventory.patch

this patch should fix issue

CoachYT1:
I continued  #318.

This PR fixes #312.

### DIFF
--- a/patches/server/0112-Fix-chest-visual-open-bug.patch
+++ b/patches/server/0112-Fix-chest-visual-open-bug.patch
@@ -1,0 +1,41 @@
+Subject: [PATCH] Fix chest visual open bug
+---
+Index: src/main/java/net/minecraft/server/PlayerConnection.java
+IDEA additional info:
+Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
+<+>UTF-8
+===================================================================
+diff --git a/src/main/java/net/minecraft/server/PlayerConnection.java b/src/main/java/net/minecraft/server/PlayerConnection.java
+--- a/src/main/java/net/minecraft/server/PlayerConnection.java	(revision c3c0231b17f661eeb1b2578be6289e2c3e60be5e)
++++ b/src/main/java/net/minecraft/server/PlayerConnection.java	(date 1748128380756)
+@@ -701,6 +701,13 @@
+         // CraftBukkit start
+         if (this.player.dead) return;
+ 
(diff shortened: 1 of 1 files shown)


uRyanxD:
Thanks for the pullrequest, but your patch is broken, maybe your git version did that? idk

patches/server/0112-Fix-chest-visual-open-bug.patch
```diff
@@ -0,0 +1,44 @@
+From 5a115d5d06dfb6ade6dbcdc92c6b13c508e8dcd2 Mon Sep 17 00:00:00 2001
+From: FabrisNick <niccolofabris05@gmail.com>
+Date: Sun, 25 May 2025 20:37:52 +0200
+Subject: [PATCH] Fix chest visual open bug
+
+---
+ .../com/hpfxd/pandaspigot/config/PandaSpigotConfig.java    | 4 ++++
```
uRyanxD, line 7: Remove these messages from modified files

patches/server/0112-Fix-chest-visual-open-bug.patch
```diff
@@ -0,0 +1,44 @@
...
+--- a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotConfig.java
++++ b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotConfig.java
+@@ -123,4 +123,8 @@ public class PandaSpigotConfig {
+     @Comment("Whether player IP addresses should be logged by the server. This does not impact\n" +
+         "the ability of plugins to log the IP addresses of players.")
+     public boolean logPlayerIpAddresses = true;
++
++    @Comment("Prevents sending interact packets while an inventory is open. This fixes the bug\n" +
++        "of chests showing as open when the aren't.")
++    public boolean dropInteractPacketsInInventory = true;
```
uRyanxD, line 22: Disable this by default, as it is not the vanilla behavior. It would be good to describe this in the comment as well and make it clear that this may cause issues with some plugin (as described in the original patch).

PR #356, opened 2025-09-23
Alternative Chunk Ticking Selection

uRyanxD:
In Minecraft 1.8 vanilla, the server selects chunks around the player (limited by the server's view distance) and stores them in a list, then ticks them. After each tick, the list is cleared.

This is a complete performance disaster. To work around this, Spigot implemented an alternative system:
It limits the ticking radius around the player to 7 and caps how many chunks are ticked per tick. Additionally, it randomly selects chunks to tick each tick based on the number of players in them, thus modifying the rate of plant growth. If a chunk is unloaded or queued for unloading, it is removed from the list and skipped.

The Spigot system is efficient, but it presents some compatibility issues with vanilla, since not all surrounding chunks are ticked.

This pull request backports the chunk selection system from Minecraft 1.9 for ticking. How it works:
- Retrieve all active PlayerChunks.
- If the chunk is null, skip to the next one.
- If the chunk needs ticking due to world generation or internal chunk systems, tick it.
- If the chunk has no players, or only spectators, and is 8 chunks away from any player (i.e., will be unloaded soon), it is skipped.

TODO:
- [x] Leave the system configurable? Since this system probably needs to use a little more memory
- [x] Add comments and cleanup
- [x] More tests


Real6dq:
nevermind*

anzz1:
Why was the tick radius hardcoded to 7 in the first place, shouldn't it respect the view distance? And the new system hardcodes it to 8, still not respecting it?
Otherwise I agree that this is a good change, the goal should for mechanics to be 1:1 to vanilla and the performance come from code optimizations and not from lessening the feature set comparing to vanilla, at least so with the default configuration.

anzz1:
I've been testing this a bit and have encountered no problems whatsoever so far.

I'm not sure if it matters but I'm using `use-async-lighting: false` which is the default and `random-light-updates: true` which is NOT the default but probably should be, considering vanilla does random light updates and without it the world is full of horrible lighting errors. This happened with the old chunk ticking selection too so I don't think the lighting system is necessarily related to these chunk ticks. Of course even better would be a system where the lights dont need to be randomly "fixed" but rather that they'd be correct in the first place, but that is probably a harder problem to fix.

patches/server/0128-Alternative-Chunk-Ticking.patch
```diff
@@ -0,0 +1,253 @@
...
++            protected Chunk computeNext() {
++                while (iterator.hasNext()) {
++                    PlayerChunk playerChunk = iterator.next();
++                    Chunk chunk = playerChunk.getChunk();
++
++                    if (chunk == null) {
++                        continue;
++                    }
++
++                    if (!chunk.u() && chunk.isDone()) { // Já tem luz e está pronto
```
anzz1: The comment should be in English

anzz1:
I was looking into the view distance thing and so confused about it until I finally figured out what's happened.

It's someone making the most unfortunate patch called `0054-Add-player-view-distance-API.patch` which apparently is an attempt to make an API where players view distance could be set/get individually.

The patch was obviously left unfinished too considering the API it left in `PandaSpigot-API\src\main\java\org\bukkit\entity\Player.java` is simply the following:

```java
public int getViewDistance()
{
    throw new UnsupportedOperationException( "Not supported yet" );
}

public void setViewDistance(int viewDistance)
{
    throw new UnsupportedOperationException( "Not supported yet" );
}
```

Of course the effects and changes of such patch would be far-reaching so apparently it was simply abandoned at some point. But the unfortunate part here is that it was never removed and it left behind the spigot configuration option "view-distance" which apparently is meant to be the default value for this player-specific view distance. This then caused confusion for other devs and now this `world.spigotConfig.viewDistance` is mistakenly used in `ActivationRange.java` and `SpawnerCreature.java`, while everything else properly uses the actual view distance the server.properties one.

What should be done first is properly revert this `0054-Add-player-view-distance-API.patch` to properly remove and clean up that unfinished player view distance nonsense, and then could get into my actual suggestion which is to make the new alternative chunk ticking system actually respect the view distance like the other entity systems already do. In fact considering the other systems do respect the view distance, i'm wondering how this current hard-coded one even works with it like the entity activate system tries activating entities by the server view distance but on the other hand the chunk tick system only ticks max 8 (the old system max 7) chunks?

Anyway then simply replacing the 
`playerChunk.hasPlayersNearby(128.0D, NON_SPECTATOR)`
with
`playerChunk.hasPlayersNearby(this.getServer().getViewDistance() * 16.0D, NON_SPECTATOR)`
should make also the chunk tick system respect the view distance and now be in line with the rest of the systems.

anzz1:
Actually now that I look into even further, there is actually *three* systems at play. The original server.properties view-distance, the Spigot view-distance, and the unfinished "player view distance" is crudely put on top of that.
Actually the spigotConfig.viewDistance is NOT from the "player view distance API", but rather `0010-View-Distance.patch`, which allows setting per-world view distances

> This commit allows the user to select per world view distances, and view distances below 3. Be wary of the issues selecting a view distance of 1 or 2 may cause!

That is maybe something that someone somewhere wants to do, but the problem is the server is now all messed up. In EntityEnderDragon, EntityLightning, EntityWither, and the old tick method the original view distance is used (except for tick its capped to 7), yet ActivationRange, SpawnerCreature, and (indirectly) PlayerChunkMap use the spigotConfig view distance.

So yes the unfinished playerviewdistance API needs to be removed, but also the per-world view distance needs to be fixed, either by removing it altogether and making everything use the vanilla server property view distance, or keeping it and fixing everything where the vanilla version is used to use the per world spigot config one. They can't be mixed like they are now that doesn't make any sense.

anzz1:
Alright the first step of removing the player view distance API is done: #405.
Now the next step is to fix everything to use the spigot config per-world view distance instead of the vanilla one.
Edit: Next step done too #406.

And to be in line with the per-world view distances, the new chunk tick method should use
`playerChunk.hasPlayersNearby(this.world.spigotConfig.viewDistance * 16.0D, NON_SPECTATOR)`

anzz1:
I went ahead and made the patch configurable:

[0131-Alternative-Chunk-Ticking-Selection.patch](https://github.com/user-attachments/files/29126796/0131-Alternative-Chunk-Ticking-Selection.patch)

This is a great improvement over the Spigot system and greatly alleviates the issue #408

Ostensibly if one were so inclined, if they have a large server and see performance degradation with large amount of players with this 1.9 system, a plugin could be made quite easily that would track TPS and if it falls below 19.5, lets say you start with view-distance of 16, it could dynamically adjust the view distance down until its down to 8, and if that is not enough then change the config variable to use the Spigot chunk system. Then at lower load it could go back to the 1.9 system with 8 view distance, going up all the way to 16 as long as TPS stays solid at 20.

Also Spigot's view-distance can be set per world (fixed in #406), as can this chunk system variable so one could use different view distances and chunk ticking systems for overworld/nether/end if they wanted.

In my opinion this patch is now ready for prime time since it's a configurable option.

anzz1:
You could push the patch I linked above its already rebased and made configurable, see https://github.com/anzz1/PandaSpigot/commit/50555777190e1ba63abcf83db6143edd6e5053c6 it applied

uRyanxD:
> You could push the patch I linked above its already rebased and made configurable, see [anzz1@5055577](https://github.com/anzz1/PandaSpigot/commit/50555777190e1ba63abcf83db6143edd6e5053c6) it applied

I'll do that soon; I'm eating right now

anzz1:
Sorry my intention was not to hurry you, I just saw you pushed new commits and was saying you needn't rebase as the posted patch is already rebased.

uRyanxD:
> Sorry my intention was not to hurry you, I just saw you pushed new commits and was saying you needn't rebase as the posted patch is already rebased.

No problem, I didn't have time to work on it yesterday; I'll review the changes today and merge them.

uRyanxD:
LGTM now

PR #361, opened 2025-10-10
Backport NetworkClient & PaperServerListPingEvent

MasterDash5:
Backports Paper's NetworkClient, which is used for obtaining a player's protocol version and virtual host.
This also lays down the groundwork to backport PaperServerListPingEvent in the future.

### DIFF
--- a/patches/server/0119-Backport-PaperServerListPingEvent.patch
+++ b/patches/server/0119-Backport-PaperServerListPingEvent.patch
@@ -332,7 +332,7 @@ index 22381439f5e2e043a5bfc18ef39553d572078180..b07e5cb71e5f5aa1fa486909a7bc0d7a
  
                          try {
 diff --git a/src/main/java/net/minecraft/server/PacketStatusListener.java b/src/main/java/net/minecraft/server/PacketStatusListener.java
-index e4bff04ac4b817f92a6a2ab6ab4e7afa6e9576bd..e5d47299bee535c586169f766b84a03fcf64d239 100644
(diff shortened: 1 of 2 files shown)


uRyanxD:
That sounds good to me, feel free to expand the PR to add the event if you want, otherwise I'll merge it tomorrow.

patches/server/0119-Backport-PaperServerListPingEvent.patch
```diff
@@ -0,0 +1,356 @@
...
++                        // PandaSpigot start - Backport PaperServerListPingEvent
++                        event = com.destroystokyo.paper.network.PaperLegacyStatusClient.processRequest(minecraftserver, inetsocketaddress, 61, null);
++
++                        if (event == null) {
++                            channelhandlercontext.close();
++                            break;
++                        }
++
++                        String s1 = String.format("\u00a71\u0000%d\u0000%s\u0000%s\u0000%d\u0000%d", event.getProtocolVersion(), event.getVersion(),
++                            com.destroystokyo.paper.network.PaperLegacyStatusClient.getMotd(event), event.getNumPlayers(), event.getMaxPlayers());
```
uRyanxD: Couldn't it just be event.getMotd()?

uRyanxD (changes requested):

patches/server/0119-Backport-PaperServerListPingEvent.patch
MasterDash5: 1.6 and older only have 1 line in the motd, so this method strips the 2nd line (anything after \n). We could merge the 2 lines into 1, but stripping the 2nd line keeps this behavior consistent with paper.

PR #390, opened 2026-04-15
Fix end portal platform

### LINKED ISSUE #389: Obsidian platform at the end not generating
Apparently, the obsidian platform in The End isn't being generated, unlike in vanilla.

Probably caused by the option to skip credits in Paper

See: https://github.com/PaperMC/Paper-archive/commit/30f02fe6e57dcbd6e001a7982c06a7728bfff58b

MasterDash5:
This changes the skip credits logic so that end portals still follow vanilla logic, and generate the end platform. This fixes #389 

This also fixes entities that travel through end portals always teleporting to the exact spawn position.

### DIFF
--- a/patches/server/0121-Make-end-overworld-spawning-more-consistent-with-van.patch
+++ b/patches/server/0121-Make-end-overworld-spawning-more-consistent-with-van.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Make end->overworld spawning more consistent with vanilla
 
 
 diff --git a/src/main/java/net/minecraft/server/PlayerList.java b/src/main/java/net/minecraft/server/PlayerList.java
-index f61dce199e2d30d425e0d8af312406cc09baf018..fab74c7512fffa92605f88e8db0b6370f46ce9b9 100644
+index f61dce199e2d30d425e0d8af312406cc09baf018..3cc5fcdf65dac0f2ecf20b464de22fe7ce3ce6ce 100644
 --- a/src/main/java/net/minecraft/server/PlayerList.java
 +++ b/src/main/java/net/minecraft/server/PlayerList.java
 @@ -680,7 +680,12 @@ public abstract class PlayerList {
@@ -22,12 +22,3 @@ index f61dce199e2d30d425e0d8af312406cc09baf018..fab74c7512fffa92605f88e8db0b6370
                  }
              } else {
                  // NORMAL <-> NETHER or NORMAL -> THE_END
-@@ -774,7 +779,7 @@ public abstract class PlayerList {
-             if (i == 1) {
-                 // use default NORMAL world spawn instead of target
-                 worldserver1 = this.server.worlds.get(0);
--                blockposition = worldserver1.getSpawn();
-+                blockposition = worldserver1.r(worldserver1.getSpawn()); // PandaSpigot - Make end->overworld spawning more consistent with vanilla
-             } else {
-                 blockposition = worldserver1.getDimensionSpawn();
-             }
--- a/patches/server/0121-Make-end-overworld-spawning-more-consistent-with-van.patch
+++ b/patches/server/0121-Make-end-overworld-spawning-more-consistent-with-van.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Make end->overworld spawning more consistent with vanilla
 
 
 diff --git a/src/main/java/net/minecraft/server/PlayerList.java b/src/main/java/net/minecraft/server/PlayerList.java
-index f61dce199e2d30d425e0d8af312406cc09baf018..3cc5fcdf65dac0f2ecf20b464de22fe7ce3ce6ce 100644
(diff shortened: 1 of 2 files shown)


patches/server/0123-Fix-end-obsidian-platform-generation.patch
```diff
@@ -0,0 +1,67 @@
...
+             this.b((Statistic) AchievementList.D);
++            this.world.kill(this); // PandaSpigot - From below. Always kill for correct end exit.
+             if (!world.paperSpigotConfig.disableEndCredits) {
+-                this.world.kill(this);
++                //this.world.kill(this); // PandaSpigot - Move above
+                 this.viewingCredits = true;
+                 this.playerConnection.sendPacket(new PacketPlayOutGameStateChange(4, 0.0F));
++            // PandaSpigot start - Manually teleport because credits are skipped.
++            } else {
++                this.server.getPlayerList().changeDimension(this, 0, TeleportCause.END_PORTAL);
```
uRyanxD, line 23: Won't that cause changeDimension to be called twice?

patches/server/0121-Make-end-overworld-spawning-more-consistent-with-van.patch
```diff
@@ -22,3 +22,12 @@ index f61dce199e2d30d425e0d8af312406cc09baf018..3cc5fcdf65dac0f2ecf20b464de22fe7
                  }
              } else {
                  // NORMAL <-> NETHER or NORMAL -> THE_END
+@@ -774,7 +779,7 @@ public abstract class PlayerList {
+             if (i == 1) {
+                 // use default NORMAL world spawn instead of target
+                 worldserver1 = this.server.worlds.get(0);
+-                blockposition = worldserver1.getSpawn();
++                blockposition = worldserver1.r(worldserver1.getSpawn()); // PandaSpigot - Make end->overworld spawning more consistent with vanilla
```
uRyanxD: Does it seem like vanilla behavior to me?

https://hub.spigotmc.org/stash/projects/SPIGOT/repos/craftbukkit/browse/nms-patches/PlayerList.patch?at=refs%2Fheads%2Fversion%2F1.8.8#680

patches/server/0123-Fix-end-obsidian-platform-generation.patch
```diff
@@ -0,0 +1,67 @@
...
++        //agent.setCanCreatePortal(cause != TeleportCause.END_PORTAL); // PaperSpigot - Configurable end credits, don't allow End Portals to create portals // PandaSpigot - The end obsidian platform is considered a "portal" and should be generated still.
+ 
+         PlayerPortalEvent event = new PlayerPortalEvent(entityplayer.getBukkitEntity(), enter, exit, agent, cause);
+         event.useTravelAgent(useTravelAgent);
+@@ -705,7 +705,7 @@ public abstract class PlayerList {
+         }
+ 
+         // PaperSpigot - Configurable end credits, if a plugin sets to use a travel agent even if the cause is an end portal, ignore it
+-        exit = cause != TeleportCause.END_PORTAL && event.useTravelAgent() ? event.getPortalTravelAgent().findOrCreate(event.getTo()) : event.getTo();
++        exit = (cause != TeleportCause.END_PORTAL || i == 1) && event.useTravelAgent() ? event.getPortalTravelAgent().findOrCreate(event.getTo()) : event.getTo(); // PandaSpigot - Use travel agent on overworld->end for platform generation
```
uRyanxD, line 64: ~If the goal here is to prevent travel agent on the return of the end, then we can simply use event.useTravelAgent(), since the code above already handles that~

nvm, this is used to ignore, even if the user wants to, it doesn't make much sense to me

patches/server/0123-Fix-end-obsidian-platform-generation.patch
MasterDash5: The player doesn't immediately teleport back to the overworld when using the end exit portal. The server set `viewingCredits` to true, and waits for the client to finish watching the credits. It does this at `PlayerConnection` line 1436. Since we skip this whole process, the teleport is included here.

patches/server/0121-Make-end-overworld-spawning-more-consistent-with-van.patch
MasterDash5: I did more testing, the entity is teleporting to the integer position of the block, instead of the center.

`World.getSpawn()` should already be moving the position of the entity to the highest block. I'm thinking Chunk heightMap is incorrect, but I'll need to investigate. I'm gonna remove this part of the PR so the end platform fix can be merged still.

patches/server/0123-Fix-end-obsidian-platform-generation.patch
MasterDash5: The travel agent is used for portal generation, which we don't want with end->overworld

uRyanxD (approved):
It seems to be working fine now, thank you

PR #400, opened 2026-05-25
Fix vanilla spawn protection (remove opslist not empty req)

### LINKED ISSUE #399: spawn-protection
Can you patch out

```
        else if (this.getConfigurationManager().getOppedPlayers().isEmpty())
        {
            return false;
        }
```

from `DedicatedServer.java:public boolean isBlockProtected(World worldIn, BlockPos pos, EntityPlayer playerIn)`
so that spawn protection always works irregardless whether there is an opped player on the server or not.
You can disable it with `spawn-protection=0` in `server.properties` anyway, so this check doesn't make much sense.

anzz1:
Removes the nonsensical requirement of having to have at least one op in the ops list for the vanilla spawn protection to work

The spawn protection can be turned off with spawn-protection=0 in server.properties anyway

Closes #399

### DIFF
--- a/patches/server/0124-Fix-vanilla-spawn-protection.patch
+++ b/patches/server/0124-Fix-vanilla-spawn-protection.patch
@@ -1,0 +1,23 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Mon, 25 May 2026 14:23:02 +0000
+Subject: [PATCH] Fix vanilla spawn protection
+
+
+diff --git a/src/main/java/net/minecraft/server/DedicatedServer.java b/src/main/java/net/minecraft/server/DedicatedServer.java
+index 8a302b2..4820219 100755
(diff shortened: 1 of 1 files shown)


uRyanxD:
I agree that this doesn't make much sense, but Mojang seems to preserve this behavior up to the current version; adding a setting to ignore this check seems like the best choice to me.
Screenshot_1

anzz1:
True, but I disagree about needing another configuration flag to control that behaviour because `spawn-protection` already exists. Setting it to 0 and setting a boolean whether the ops requirement is lifted would be two configuration options for the same thing which is unnecessary and unintuitive. Common design logic says there should ever be one option to control one thing.

The configuration is already confusing due to historical baggage having `bukkit.yml`, `pandaspigot.yml`, `paper.yml`, `spigot.yml` so the config is illogically spread over multiple files, and sometimes same thing is in two places like `view-distance` in both `server.properties` and `spigot.yml`, let's not add to that confusion.

uRyanxD (approved):

PR #401, opened 2026-05-30
Fix VehicleDamageEvent API

anzz1:
Make the damage adjustable for EntityBoat as it is for EntityMinecartAbstract


anzz1:
As a sidenote, attached is a plugin that makes boats much more usable in 1.8.9, making them take 50% less damage and not colliding with players and getting yeeted when nobody is riding. The source is inside the jar file. Requires to have this change merged for it to work.

[ToughBoats.zip](https://github.com/user-attachments/files/28686871/ToughBoats.zip)

uRyanxD (approved):
LGTM, thank you, it just needs to be rebased.

anzz1:
It is done.

PR #403, opened 2026-06-07
Remove old tile entity on setTypeIdAndData

MasterDash5:
Bukkit forgot to handle tile entities with the `Block.setType` API. This can result in the server attempting to use the wrong tile entity data for a block. This doesn't crash the server as Bukkit has a fail-safe for this, but it does log errors.

This notably can happen with WorldEdit. Using //set for example to replace a chest with a dropper will trigger this issue.


uRyanxD (approved):
LGTM, just needs to be rebased.

anzz1:
@uRyanxD

Btw, as long as the OP didn't uncheck the "Allow edits by maintainers", maintainers like you have edit access to the PR branch, useful in cases like this where the OP is not active and the patch needs to be renamed to update its patch number.

Then the PR can simply be rebased/squashed and committed from the GitHub UI without having to go actually rebase the branch manually as long as "Allow rebase merging" and "Allow squash merging" are enabled in the repo settings.

In my own repos I actually disallow ugly regular merge commits altogether and only allow rebase and squash merges as per the instructions in [GitHub New Repo 101](https://gist.github.com/anzz1/b2d92194e30768a1c340e142bb1ae542)

uRyanxD:
> @uRyanxD
> 
> Btw, as long as the OP didn't uncheck the "Allow edits by maintainers", maintainers like you have edit access to the PR branch, useful in cases like this where the OP is not active and the patch needs to be renamed to update its patch number.
> 
> Then the PR can simply be rebased/squashed and committed from the GitHub UI without having to go actually rebase the branch manually as long as "Allow rebase merging" and "Allow squash merging" are enabled in the repo settings.
> 
> In my own repos I actually disallow ugly regular merge commits altogether and only allow rebase and squash merges as per the instructions in [GitHub New Repo 101](https://gist.github.com/anzz1/b2d92194e30768a1c340e142bb1ae542)

I am not the owner of the repository; I don't have access.

PR #405, opened 2026-06-18
[WIP] Finish the unfinished player view distance API so its used everywhere

anzz1:
This reverts the `0054-Add-player-view-distance-API.patch` from Spigot. The API was never finished and was later removed on newer Spigot as a system for per-player view distance would be unnecessarily complex and afaik the client doesn't transmit its preferred render distance anyway. Clean up the vestigial remains of this never-finished system.

The API side Player.getViewDistance() and Player.setViewDistance() which just throw UnsupportedOperationException anyway still stays to have 1:1 api compat with the original 1.8.8 spigot.

### DIFF
--- a/patches/server/0128-Remove-unfinished-player-view-distance-API.patch
+++ b/patches/server/0128-Remove-unfinished-player-view-distance-API.patch
@@ -1,0 +1,136 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Thu, 18 Jun 2026 10:09:40 +0000
+Subject: [PATCH] Remove unfinished player view distance API
+
+
+diff --git a/src/main/java/net/minecraft/server/EntityPlayer.java b/src/main/java/net/minecraft/server/EntityPlayer.java
+index 6f804d5b1eaa085366b5213a98bc412cada1f078..79a88319c7c89dd02e053b9b363ff8e630f96bf7 100644
+--- a/src/main/java/net/minecraft/server/EntityPlayer.java
++++ b/src/main/java/net/minecraft/server/EntityPlayer.java
+@@ -65,7 +65,6 @@ public class EntityPlayer extends EntityHuman implements ICrafting {
+     // CraftBukkit end
+     // Spigot start
+     public boolean collidesWithEntities = true;
+-    public int viewDistance; // PaperSpigot - Player view distance API
+     private int containerUpdateDelay; // PaperSpigot
+ 
+     @Override
+@@ -83,7 +82,6 @@ public class EntityPlayer extends EntityHuman implements ICrafting {
+ 
+     public EntityPlayer(MinecraftServer minecraftserver, WorldServer worldserver, GameProfile gameprofile, PlayerInteractManager playerinteractmanager) {
+         super(worldserver, gameprofile);
+-        this.viewDistance = world.spigotConfig.viewDistance; // PaperSpigot - Player view distance API
+         playerinteractmanager.player = this;
+         this.playerInteractManager = playerinteractmanager;
+         BlockPosition blockposition = getSpawnPoint(minecraftserver, worldserver); // PandaSpigot - Split spawn point logic
+diff --git a/src/main/java/net/minecraft/server/PlayerChunkMap.java b/src/main/java/net/minecraft/server/PlayerChunkMap.java
+index 1c9dbe212cfbeb571f1e09f224fb7ed905dd475b..e1880604533f42c050b2bd0195f9f46c93409cc4 100644
+--- a/src/main/java/net/minecraft/server/PlayerChunkMap.java
++++ b/src/main/java/net/minecraft/server/PlayerChunkMap.java
+@@ -129,10 +129,8 @@ public class PlayerChunkMap {
+         // CraftBukkit start - Load nearby chunks first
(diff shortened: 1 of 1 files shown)


kalibrieren (changes requested):
Signaling that this API was removed in PandaSpigot would be better? The message "Not supported yet" isn't too clear.

anzz1:
The API was not removed it never existed in the first place, I'm just cleaning up the scaffolding. The "Not supported yet" is by who originally started it and I simply didn't touch it. So that message should be changed to remove the "yet" part?

edit: i went to look into the APIs thinking of removing the "yet" but every other non existent API also prints "Not supported yet" even if there is probably no plans to support most of them so I think it's fine keeping it the way it is.

anzz1:
The build works fine, it passed the test before I renamed the patch. The failing check is just some unrelated server issue

> Task :pandaspigot-api:javadoc
error: Error fetching URL: https://javadoc.io/doc/org.slf4j/slf4j-api/1.7.35/ (java.io.IOException: Server returned HTTP response code: 522 for URL: https://javadoc.io/doc/org.slf4j/slf4j-api/1.7.35/package-list)

uRyanxD:
1 - This API isn't from Spigot but rather from Paper (the patch folder name is stupid; they use the downstream's name instead of their own);
2 - The API exists, but the exception isn't thrown because the API is a "placeholder" and the code that actually executes is the method in CraftServer—I can't quite explain how it works;
3 - It's likely not working quite right.
Check out this Paper 1.9 commit for reference: https://github.com/PaperMC/Paper-archive/commit/2cbd400e174c0bdfa61472f2150efe03aeaba2d2

kalibrieren:
> 1 - This API isn't from Spigot but rather from Paper (the patch folder name is stupid; they use the downstream's name instead of their own);
> 2 - The API exists, but the exception isn't thrown because the API is a "placeholder" and the code that actually executes is the method in CraftServer—I can't quite explain how it works;
> 3 - It's likely not working quite right.
> Check out this Paper 1.9 commit for reference: https://github.com/PaperMC/Paper-archive/commit/2cbd400e174c0bdfa61472f2150efe03aeaba2d2

I remember this API being removed around 1.19. Removing it shouldn't be a problem?
Or just go the route of trying to fix it?

anzz1:
> The API exists, but the exception isn't thrown because the API is a "placeholder" and the code that actually executes is the method in CraftServer—I can't quite explain how it works;

Interesting, there is indeed so many layers of systems on top of each other here that it gets often very confusing.

> I remember this API being removed around 1.19. Removing it shouldn't be a problem?
Or just go the route of trying to fix it?

Yeah, I don't think this API makes much sense and removing it cleans up the codebase nicely, at least for a little bit. Actually fixing it piping it to all the systems that use view distance would be quite messy as those systems were never designed for such a thing in the first place.
To make it properly it'd either need redesigning of the systems, or hacking it together with messy hacks.
Rather just remove the broken mess, and I can't really think of any reason why someone would even need it. It can always be re-implemented, properly this time, at a later time.

Beaness:
The advantage of keeping a per player viewdistance is that the server can send less chunks for players with lower view distance than server view distance, that's also what the updateViewDistance kind of does.

anzz1:
> The advantage of keeping a per player viewdistance is that the server can send less chunks for players with lower view distance than server view distance, that's also what the updateViewDistance kind of does.

Oh well that actually does change everything then and absolutely makes this worthwhile to implement and fix. I can't find anything in this damn codebase where nothing is named in any sensible way so I'll refer to the proper SRG names here as follows. I simply don't understand why Spigot couldn't have gone the route of MCP where you can work with proper names of everything after a de-obfuscation step, and then for reflection and stable function naming purposes re-obfuscated back on compilation.

So anyway, I always thought that this client sending its render distance is a much newer feature and thus the setPlayerViewDistance would have been only confined to usage by plugins through the API as there would be no way to retrieve this information from the player. **But the client does actually send its view distance in the C15PacketClientSettings packet**, but I never realized this because that's the only thing that happens it was left completely unfinished from mojang's side. As in when the packet arrives and is deserialized to NetHandlerPlayServer, the contents get passed on to EntityPlayerMP.handleClientSettings(), but the render distance part is left completely unused, not stored anywhere, EntityPlayerMP doesn't even have such a field nor is it ever used later in determining what chunks are sent to players, only the view-distance from server.properties is.

From what I can tell even though I still can't understand everything that is happening in spigot, the player's view distance is set only in the ctor using the world.spigotConfig.viewDistance, well at least it uses the spigot config's per-world view distance, but then its never updated anymore and is functionally same as not having it at all.

So to actually make it work, the following should be done:
- in the EntityPlayerMP.handleClientSettings(), which is called `EntityPlayer.a(PacketPlayInSettings packetplayinsettings)` in spigot I believe, the players viewDistance must be updated if its lower than the server's world one
- when player changes between worlds, the view distance must be updated again to either match the player's setting or the world's, whichever is lower. so giving the player another variable that keeps whatever the player sent at its setting in addition to the what is the current one should be used
- all this is connected including the PR #406 which should be closed and all the fixes to various overlapping systems merged into one

The ultimate source of confusion here is naming this as an "API" as if its something to be used by plugins, when this whole thing is not about any API nor plugins using it but setting it according to what the client sends.

I'll now leave this hanging as a draft and when I time completely rewrite the patch as this whole system ties in with #406 it all should be fixed in one go include the things from that one too.

PR #406, opened 2026-06-18
Fix Spigot per-world view distance

anzz1:
`view-distance` can be configured per-world in `spigot.yml` instead of using
the value from `server.properties` globally.
However some of the systems still used the original global value instead
of the Spigot one. This patch fixes those inconsistencies.

Also backport "default" view distance from modern Spigot.
This will not affect current configurations but simply make new
configurations use the global default from `server.properties`.
(Setting `view-distance` option to an invalid integer, which "default" is,
will make it use the global default, this was already the case)

> Note: As of 1.14.4, the *global* option is set to "default" by default for new installs & upgrades from older versions. When set to default, the server reads the view-distance set in server.properties. This setting can now be used for per-world view distances.
https://www.spigotmc.org/wiki/spigot-configuration/

### DIFF
--- a/patches/server/0129-Fix-Spigot-per-world-view-distance.patch
+++ b/patches/server/0129-Fix-Spigot-per-world-view-distance.patch
@@ -1,0 +1,90 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Fri, 19 Jun 2026 03:18:09 +0000
+Subject: [PATCH] Fix Spigot per-world view distance
+
+View-distance can be configured per-world in spigot.yml instead of using
+the value from server.properties globally.
+However some of the systems still used the original global value instead
+of the Spigot one. This patch fixes those inconsistencies.
+
+Also backport "default" view distance from modern Spigot.
+This will not affect current configurations but simply make new
+configurations use the global default from server.properties.
+(Setting view-distance option to an invalid integer, which "default" is,
+will make it use the global default, this was already the case)
+
+diff --git a/src/main/java/net/minecraft/server/EntityEnderDragon.java b/src/main/java/net/minecraft/server/EntityEnderDragon.java
+index d246ae749ea49be9e07ff5ac6abd3bbf14782a6b..8a5025472dc039256740bb941e3fd0c5e7db684d 100644
+--- a/src/main/java/net/minecraft/server/EntityEnderDragon.java
++++ b/src/main/java/net/minecraft/server/EntityEnderDragon.java
+@@ -573,7 +573,7 @@ public class EntityEnderDragon extends EntityInsentient implements IComplex, IMo
+             if (this.by == 1) {
+                 // CraftBukkit start - Use relative location for far away sounds
+                 // this.world.a(1018, new BlockPosition(this), 0);
+-                int viewDistance = ((WorldServer) this.world).getServer().getViewDistance() * 16;
++                int viewDistance = this.world.spigotConfig.viewDistance * 16; // PandaSpigot - Fix Spigot per-world view distance
+                 for (EntityPlayer player : (List<EntityPlayer>) MinecraftServer.getServer().getPlayerList().players) {
+                     double deltaX = this.locX - player.locX;
+                     double deltaZ = this.locZ - player.locZ;
+diff --git a/src/main/java/net/minecraft/server/EntityLightning.java b/src/main/java/net/minecraft/server/EntityLightning.java
+index 286fdef9aa1cd9c0d3c399d483b7cd4a26d7aa63..9cfc03e032d1e163438a4353db9d4cae9d1dbe6a 100644
+--- a/src/main/java/net/minecraft/server/EntityLightning.java
++++ b/src/main/java/net/minecraft/server/EntityLightning.java
+@@ -70,7 +70,7 @@ public class EntityLightning extends EntityWeather {
+             // CraftBukkit start - Use relative location for far away sounds
+             //this.world.makeSound(this.locX, this.locY, this.locZ, "ambient.weather.thunder", 10000.0F, 0.8F + this.random.nextFloat() * 0.2F);
+             float pitch = 0.8F + this.random.nextFloat() * 0.2F;
+-            int viewDistance = ((WorldServer) this.world).getServer().getViewDistance() * 16;
++            int viewDistance = this.world.spigotConfig.viewDistance * 16; // PandaSpigot - Fix Spigot per-world view distance
+             for (EntityPlayer player : (List<EntityPlayer>) (List) this.world.players) {
+                 double deltaX = this.locX - player.locX;
+                 double deltaZ = this.locZ - player.locZ;
+diff --git a/src/main/java/net/minecraft/server/EntityWither.java b/src/main/java/net/minecraft/server/EntityWither.java
+index fb19bad3c01b03991f9dbaff6041acbccda3c8b3..7998cd4dd0546f36a41712b7d71ecfe34b873b8f 100644
+--- a/src/main/java/net/minecraft/server/EntityWither.java
++++ b/src/main/java/net/minecraft/server/EntityWither.java
+@@ -186,7 +186,7 @@ public class EntityWither extends EntityMonster implements IRangedEntity {
+ 
+                 // CraftBukkit start - Use relative location for far away sounds
+                 // this.world.a(1013, new BlockPosition(this), 0);
+-                int viewDistance = ((WorldServer) this.world).getServer().getViewDistance() * 16;
++                int viewDistance = this.world.spigotConfig.viewDistance * 16; // PandaSpigot - Fix Spigot per-world view distance
+                 for (EntityPlayer player : (List<EntityPlayer>) MinecraftServer.getServer().getPlayerList().players) {
+                     double deltaX = this.locX - player.locX;
+                     double deltaZ = this.locZ - player.locZ;
+diff --git a/src/main/java/net/minecraft/server/World.java b/src/main/java/net/minecraft/server/World.java
+index 9b86cdd3b83ae8b0c040dc9d250b54684d0b7510..90c7a6f4777d2fa2bb08b539be59ead6772572d1 100644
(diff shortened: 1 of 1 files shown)


uRyanxD:
I believe the entity tracker (in the constructor) is also using the value from server.properties instead of the per-world value from Spigot.

anzz1:
That indeed is true, I did miss that one. Damn it's hard when these functions are called .a(), .b(), c.() and when that's not enough same class like PlayerList has 10 different .a() functions only differentiated by their arguments.

It would be so much easier to be able to use proper mappings but for some reason all that time ago when Bukkit that started it all was invented (?) these namings were locked in and now we have to live with them.

Anyway here's the functions involved that I traced back with their "proper" mapped names

EntityTracker this.e -> PlayerList.d() -> PlayerChunkMap.getFurthestViewableBlock(PlayerList.s()) -> PlayerList.s() -> PlayerList.r -> PlayerList.a(view-distance) -> DedicatedPlayerList ctor

EntityTracker.maxTrackingDistanceThreshold -> ServerConfigurationManager.getEntityViewDistance() -> PlayerManager.getFurthestViewableBlock(ServerConfigurationManager.getViewDistance()) -> ServerConfigurationManager.getViewDistance() -> ServerConfigurationManager.setViewDistance(int distance) -> DedicatedPlayerList ctor

PlayerList = ServerConfigurationManager
PlayerChunkMap = PlayerManager

PlayerList.a() = ServerConfigurationManager.setViewDistance(int distance)
PlayerList.r = ServerConfigurationManager.this.viewDistance
PlayerList.s() = ServerConfigurationManager.getViewDistance()
PlayerList.d() = ServerConfigurationManager.getEntityViewDistance()
EntityTracker.e = EntityTracker.maxTrackingDistanceThreshold

WorldServer.q() -> this.server.getPlayerList().s()
WorldServer.getRenderDistanceChunks() -> this.mcServer.getConfigurationManager().getViewDistance();

World.D() uses WorldServer.q()
World.setActivePlayerChunksAndCheckLight() uses WorldServer.getRenderDistanceChunks()

So from what I see the PlayerList aka ServerConfigurationManager being a singleton, it needn't be touched, but rather change what uses them further down the line in the worlds.

Changing WorldServer.q() to return the spigot view distance takes care of World.D()
For EntityTracker there is two choices, either in the ctor change this.e to the spigotConfig viewDistance, or in the addEntity the only function that uses it replace it altogether.
EntityTracker is created in WorldServer after it calls super() i.e. World which constructs and inits SpigotWorldConfig, so the value is initialized at that point, so the EntityTracker ctor would be safe.

However if one were to want to make something like dynamic view distance then it'd be better done inside the addEntity, but currently there is one other place where a ctor sets the value once that is PlayerChunkMap. But on the other hand, considering neither the spigot system or the new 1.9 chunk tick system use the "players watching chunk" list anymore (?), this might be meaningless? You can probably better understand this considering you just worked with the chunk tick system.
It also seems that even mojang didn't really know what they were doing since from what I understand, the PlayerChunkMap ctor calls a() aka setPlayerViewRadius using the server.properties viewDistance value, which Spigot changed to use its own viewDistance at some point, but then the DedicatedPlayerList through PlayerList/ServerConfigurationManager in its ctor iterates through the WorldServer's PlayerChunkMap's and calls a() again overwriting the value. In vanilla it means it gets called twice with the same value from different places, but in Spigot considering they never changed the other source the viewDistance then got immediately overwritten. If I'm looking all this right, that is.

So for now, to make everything actually use the spigot's viewdistance, what needs to be still changed is
WorldServer.q()
EntityTracker this.e
remove the second call in DedicatedPlayerList ctor to PlayerList.a()

Further on, if there is want to make it possible to have the viewDistance dynamic, the last important piece here is whether the PlayerChunkMap.a() is actually important, does either of the chunk tick systems use it?
If it is then it'd need a new API call to set a new view distance which would set the both the config value and then call PlayerList.a()
But on the other hand, if the "chunk watching" map the result of PlayerList.a() isn't used at all anymore, could the whole thing just be commented out and performance improved?

uRyanxD:
> That indeed is true, I did miss that one. Damn it's hard when these functions are called .a(), .b(), c.() and when that's not enough same class like PlayerList has 10 different .a() functions only differentiated by their arguments.
> 
> It would be so much easier to be able to use proper mappings but for some reason all that time ago when Bukkit that started it all was invented (?) these namings were locked in and now we have to live with them.
> 
> Anyway here's the functions involved that I traced back with their "proper" mapped names
> 
> EntityTracker this.e -> PlayerList.d() -> PlayerChunkMap.getFurthestViewableBlock(PlayerList.s()) -> PlayerList.s() -> PlayerList.r -> PlayerList.a(view-distance) -> DedicatedPlayerList ctor
> 
> EntityTracker.maxTrackingDistanceThreshold -> ServerConfigurationManager.getEntityViewDistance() -> PlayerManager.getFurthestViewableBlock(ServerConfigurationManager.getViewDistance()) -> ServerConfigurationManager.getViewDistance() -> ServerConfigurationManager.setViewDistance(int distance) -> DedicatedPlayerList ctor
> 
> PlayerList = ServerConfigurationManager PlayerChunkMap = PlayerManager
> 
> PlayerList.a() = ServerConfigurationManager.setViewDistance(int distance) PlayerList.r = ServerConfigurationManager.this.viewDistance PlayerList.s() = ServerConfigurationManager.getViewDistance() PlayerList.d() = ServerConfigurationManager.getEntityViewDistance() EntityTracker.e = EntityTracker.maxTrackingDistanceThreshold
> 
> WorldServer.q() -> this.server.getPlayerList().s() WorldServer.getRenderDistanceChunks() -> this.mcServer.getConfigurationManager().getViewDistance();
> 
> World.D() uses WorldServer.q() World.setActivePlayerChunksAndCheckLight() uses WorldServer.getRenderDistanceChunks()
> 
> So from what I see the PlayerList aka ServerConfigurationManager being a singleton, it needn't be touched, but rather change what uses them further down the line in the worlds.
> 
> Changing WorldServer.q() to return the spigot view distance takes care of World.D() For EntityTracker there is two choices, either in the ctor change this.e to the spigotConfig viewDistance, or in the addEntity the only function that uses it replace it altogether. EntityTracker is created in WorldServer after it calls super() i.e. World which constructs and inits SpigotWorldConfig, so the value is initialized at that point, so the EntityTracker ctor would be safe.
> 
> However if one were to want to make something like dynamic view distance then it'd be better done inside the addEntity, but currently there is one other place where a ctor sets the value once that is PlayerChunkMap. But on the other hand, considering neither the spigot system or the new 1.9 chunk tick system use the "players watching chunk" list anymore (?), this might be meaningless? You can probably better understand this considering you just worked with the chunk tick system. It also seems that even mojang didn't really know what they were doing since from what I understand, the PlayerChunkMap ctor calls a() aka setPlayerViewRadius using the server.properties viewDistance value, which Spigot changed to use its own viewDistance at some point, but then the DedicatedPlayerList through PlayerList/ServerConfigurationManager in its ctor iterates through the WorldServer's PlayerChunkMap's and calls a() again overwriting the value. In vanilla it means it gets called twice with the same value from different places, but in Spigot considering they never changed the other source the viewDistance then got immediately overwritten. If I'm looking all this right, that is.
> 
> So for now, to make everything actually use the spigot's viewdistance, what needs to be still changed is WorldServer.q() EntityTracker this.e remove the second call in DedicatedPlayerList ctor to PlayerList.a()
> 
> Further on, if there is want to make it possible to have the viewDistance dynamic, the last important piece here is whether the PlayerChunkMap.a() is actually important, does either of the chunk tick systems use it? If it is then it'd need a new API call to set a new view distance which would set the both the config value and then call PlayerList.a() But on the other hand, if the "chunk watching" map the result of PlayerList.a() isn't used at all anymore, could the whole thing just be commented out and performance improved?

In my opinion, dynamic view distance is a bit out of scope for Panda —perhaps for a hard fork derived from it.

uRyanxD:
It’s not directly related, but:

If you have any questions about development for Panda/Minecraft, you can join our Discord (https://discord.gg/XwfwgkGM), where contributors usually interact whenever questions arise

anzz1:
> In my opinion, dynamic view distance is a bit out of scope for Panda —perhaps for a hard fork derived from it.

I agree, I wasn't thinking of adding it to the server itself, but rather have the possibility for a plugin to do it, as far as I can tell the pieces are already there the view distance simply needs to be exposed. But on the other hand if no one requests such a thing, its not necessary. Complex systems like that should be the purview of plugins anyway to keep the server core as clean and thus maintainable as possible. Codebase rot is the worst thing that can be done to a longevity of a project.

> If you have any questions about development for Panda/Minecraft, you can join our Discord (https://discord.gg/XwfwgkGM), where contributors usually interact whenever questions arise

[rant]
That is a good option to have, however I have love-hate relationship with discord, I particularly don't like the ephemerality of it, as in now that everything is in discord, the search is crap and people end up asking same questions all over, and unlike the discussion boards of old the contents are not indexable and searchable by search engines so all the information there is good as lost. I think this has been a great contributor of the "why I can't find anything in google anymore" because info is not getting posted in publicly indexable internet anymore. I can't count the times I've resolved some issue finding some forum post from 2005, but now the last decade feels like a lost decade when it comes to information. Easier to find detailed information on an obscure NES game from the 80s than some contemporary AAA game title.
[/rant]

That being said, I'll be joining it anyway :sweat_smile:

anzz1:
[New information has come to light](https://github.com/hpfxd/PandaSpigot/pull/405#issuecomment-4763076557) and the client does actually send it's render distance, it just isn't used but absolutely should as in case its lower than the server's (world's) one, the server can update and send less chunks accordingly which can be a huge free performance increase when players use lower render distance values.
Hence this whole thing needs to be fixed in concert with #405 as its all connected, the per-player view distance needs to not be removed but be fixed and used, and the overlapping systems needs to use that value instead, closing this PR.

kalibrieren:
should probably just be temporarily merged for now until per-player view distance is implemented

anzz1:
This isn't complete as mentioned in earlier posts the entity tracker and active player chunks system are still missing, and it makes sense to complete the work in tandem with the per-player view distance as the systems are intrinsically linked.

PR #407, opened 2026-06-18
Fix MC-87 map scaling/cloning issues

anzz1:
Backport MC-87 fix from 1.11
Fixes map not scaling when shift-clicking the result from workbench
And NBT tags not cloning when cloning a map

https://bugs.mojang.com/browse/MC/issues/MC-87

https://minecraft.wiki/w/Map?oldid=940311
> Please note: shift-clicking to retrieve a zoomed-out map will cause the eight extra sheets of paper to be consumed, but the map will be unchanged. The data from the old map will not be carried over to the new one.
> This is Minecraft Bug [MC-87](https://bugs.mojang.com/browse/MC-87).

Fixed in [Java Edition 16w32a](https://minecraft.wiki/w/Java_Edition_16w32a) (1.11 snapshot)
And now in PandaSpigot too :smile:


uRyanxD (approved):
LGTM, thank you.

PR #409, opened 2026-06-19
Entity tracking range vanilla option

anzz1:
Allow setting 'entity-tracking-range' to -1 to use vanilla entity tracking range values
Also functions as a sanity check for negative values

### DIFF
--- a/patches/server/0129-Entity-tracking-range-vanilla-option.patch
+++ b/patches/server/0129-Entity-tracking-range-vanilla-option.patch
@@ -1,0 +1,62 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Fri, 19 Jun 2026 11:21:11 +0000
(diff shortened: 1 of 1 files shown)


kalibrieren:
afaik `defaultRange` is always `0`?

anzz1:
> afaik `defaultRange` is always `0`?

No, the `defaultRange` comes from `EntityTracker.addEntity`, which is called by `EntityTracker.track` that has the table of vanilla entity tracking ranges for every entity type which is how it works without the Spigot changes.

There seems to have been an idea to use the defaultRange since it was baked into the function but then it was never used. The code quality of lot of the Spigot code isn't great, I do understand now that when they say Spigot was created because "Bukkit wasn't merging new changes fast enough", seems to me that the original Bukkit team were simply doing due diligence when it comes to code quality, while Spigot, to put it bluntly, was not.

uRyanxD (approved):
LGTM; it just needs to be rebased and the patch number corrected.

anzz1:
It is done.

PR #411, opened 2026-06-21
Check randomLightUpdates

uRyanxD:
When the chunk is populated, it sets:
done = true
lit = true
However, under certain conditions, `lit` might be set to `false` and attempt to update again on the next tick—though this only happens if `randomLightUpdates` is enabled.

In theory, the best approach here would be to keep `randomLightUpdates` enabled at all times (otherwise, theoretically, the lighting could never be populated correctly if it failed the first time).

However, I believe that would cause significant performance issues.

---

Did I mention I hate the chunk/lighting system in this version? 😂

### DIFF
--- a/patches/server/0128-Alternative-Chunk-Ticking-Selection.patch
+++ b/patches/server/0128-Alternative-Chunk-Ticking-Selection.patch
@@ -61,7 +61,7 @@ index 652c07975c7b3a3c5e3dc68cb60ca5fc52e1848c..d93670b5fd2ee58a7b1d4ccd2479cc43
          return (this.x << 4) + 8;
      }
 diff --git a/src/main/java/net/minecraft/server/PlayerChunkMap.java b/src/main/java/net/minecraft/server/PlayerChunkMap.java
-index 1c9dbe212cfbeb571f1e09f224fb7ed905dd475b..4e5139ff054fdec1f1c603f720b37a74f572196d 100644
+index 1c9dbe212cfbeb571f1e09f224fb7ed905dd475b..6a489b4514d565b47d930a6ed1abf94cd863cf54 100644
 --- a/src/main/java/net/minecraft/server/PlayerChunkMap.java
 +++ b/src/main/java/net/minecraft/server/PlayerChunkMap.java
 @@ -1,5 +1,9 @@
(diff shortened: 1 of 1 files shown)


anzz1:
Maybe in a large server the random light updates could cause performance bottleneck, but don't know how to test and simulate for let's say 100 of players. Probably that's why it's on by default even though that's not vanilla and it causes horrible lighting errors everywhere. By enabling it most of all the lighting errors get (eventually) fixed.

Well actually, now that I think of it, the light updates are just random and only increase by active chunk count, not player count, correct? I guess increasing the amount of active chunks like even just by stopping them from ever de-activating and then either programmatically or by flying around activating more chunks and checking the timings seeing how much the light updates really affect performance.

I pretty much consider a vanilla world unplayable with the random light updates off there are so much lighting errors everywhere, but this option obviously is good for custom worlds where lighting is not an issue and where light updates would be unnecessary work. This does however bring into question whether random light updates default value should be false considering the goal is to be mechanics-wise like vanilla as closely as possible with the default config.

uRyanxD:
> Maybe in a large server the random light updates could cause performance bottleneck, but don't know how to test and simulate for let's say 100 of players. Probably that's why it's on by default even though that's not vanilla and it causes horrible lighting errors everywhere. By enabling it most of all the lighting errors get (eventually) fixed.
> 
> Well actually, now that I think of it, the light updates are just random and only increase by active chunk count, not player count, correct? I guess increasing the amount of active chunks like even just by stopping them from ever de-activating and then either programmatically or by flying around activating more chunks and checking the timings seeing how much the light updates really affect performance.
> 
> I pretty much consider a vanilla world unplayable with the random light updates off there are so much lighting errors everywhere, but this option obviously is good for custom worlds where lighting is not an issue and where light updates would be unnecessary work. This does however bring into question whether random light updates default value should be false considering the goal is to be mechanics-wise like vanilla as closely as possible with the default config.

Yes. There is also something interesting we could do:

Leaf created a version of Starlight for 1.7.3 (I don't know the reason), but theoretically, we could adapt it for 1.8.8; unfortunately, I don't have the time for that.

Here is the patch, in case you or another external contributor wants to port it:
https://github.com/Spottedleaf/canyon/blob/master/CraftBukkit-Patches/0031-Starlight.patch

anzz1:
Oh yeah porting Starlight definitely looks interesting, that'd fix the lighting issues the correctness AND performance, fantastic. No need for random light updates either, the whole mechanism of "we'll make broken lighting first and then randomly update to fix them" is so silly in the first place.

Anyway this patch seems correct as it follows the same logic that spigot use in its chunk ticking, that is it uses the `randomLightUpdates` value for both the random light updates in `World.D()` aka `World.setActivePlayerChunksAndCheckLight()` and the relighting check in `Chunk.b()` aka `Chunk.onTick()` : 
```java
if (!this.lit && this.done && this.world.spigotConfig.randomLightUpdates) { // Spigot - also use random light updates setting to determine if we should relight
```
So with this patch the lighting update behaviour is consistent with both chunk ticking systems.

Issue #190, opened 2024-03-09
 Tool break sound/animation


ghost:
Anyways, it's recently been brought to my attention that there is a bug in the game that prevents the particle animation and sound effect of tools breaking in your hand. So, I scrounged this up as a fix. Seems to work ok, but I'm sure there's some things wrong with it (edge cases, etc etc). So, I decided to just post it here see what your thoughts were, or if anyone can spot any issues, or maybe 1-up it.

Details: https://www.spigotmc.org/threads/tool-break-sound-animation.297264/

PR #412, opened 2026-06-21
[WIP] Equipment break effects fixes

anzz1:
Half of the bug is client-side so this cannot entirely be fixed. Item breaking happens client-side, this is usually fine except with Unbreaking which introduces RNG, and now client and server no longer agree on whether the item was broken.

This patch makes item breaking effects, sound and particles, happen on server-side. So no more silent breakdowns, client will always hear the item breaking. However the client-side logic cannot be fixed, so if client rolls break and server does not, client will stil hear this false item break. If the client and server roll break at the same time, client will hear the sound twice, but this works out fine.

Importantly, an inventory update is now sent every time there is a chance that the Unbreaking item was broken. This syncs up the server and client, no more ghost items when item breaks server-side but not on client-side.

Edit: as discussed below, broadened scope of this PR to include all breaking effects, not just unbreaking. This currently fixes the unbreaking by brute-force method as its the only thing possible that fixes the unbreaking, but a non-bruteforce method could be possible for the non-unbreaking (non-RNG) sounds&particles.

Relevant bugs:
[MC-1040](https://bugs.mojang.com/browse/MC/issues/MC-1040) 
[MC-2518](https://bugs.mojang.com/browse/MC/issues/MC-2518)

### DIFF
--- a/patches/server/0999-Unbreaking-Break-Effect-Fix.patch
+++ b/patches/server/0999-Unbreaking-Break-Effect-Fix.patch
@@ -1,0 +1,93 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Fri, 26 Jun 2026 12:33:53 +0000
+Subject: [PATCH] Unbreaking Break Effect Fix
+
+Half of the bug is client-side so this cannot entirely be fixed.
+Item breaking happens client-side, this is usually fine except with
+Unbreaking which introduces RNG, and now client and server no longer
+agree on whether the item was broken.
+
+This patch makes item breaking effects, sound and particles, happen
+on server-side. So no more silent breakdowns, client will always hear
+the item breaking. However the client-side logic cannot be fixed, so
+if client rolls break and server does not, client will stil hear
+this false item break. If the client and server roll break at the
+same time, client will hear the sound twice, but this works out fine.
+
+Importantly, an inventory update is now sent every time there is a
+chance that the Unbreaking item was broken. This syncs up the server
+and client, no more ghost items when item breaks server-side but not
+on client-side.
+
+diff --git a/src/main/java/net/minecraft/server/EntityLiving.java b/src/main/java/net/minecraft/server/EntityLiving.java
+index 32752dfad24bc7883b1ab52e29834f1aa6fa60c6..c307907ba9fb8ab2997da848ac56d2aaeadc1cc0 100644
+--- a/src/main/java/net/minecraft/server/EntityLiving.java
++++ b/src/main/java/net/minecraft/server/EntityLiving.java
+@@ -832,9 +832,31 @@ public abstract class EntityLiving extends Entity {
+         }
+     }
+ 
++    // PandaSpigot start - Unbreaking Break Effect Fix
++    public void renderBrokenItemStack(ItemStack itemstack, boolean broadcast) {
++        if (broadcast) {
++            super.makeSound("random.break", 0.8F, 0.8F + this.world.random.nextFloat() * 0.4F);
++            Vec3D vec3d = new Vec3D(((double) this.random.nextFloat() - 0.5D) * 0.1D, Math.random() * 0.1D + 0.1D, 0.0D);
++            vec3d = vec3d.a(-this.pitch * 3.1415927F / 180.0F);
++            vec3d = vec3d.b(-this.yaw * 3.1415927F / 180.0F);
++            double d0 = (double) (-this.random.nextFloat()) * 0.6D - 0.3D;
++
++            Vec3D vec3d1 = new Vec3D(((double) this.random.nextFloat() - 0.5D) * 0.3D, d0, 0.6D);
++            vec3d1 = vec3d1.a(-this.pitch * 3.1415927F / 180.0F);
++            vec3d1 = vec3d1.b(-this.yaw * 3.1415927F / 180.0F);
++            vec3d1 = vec3d1.add(this.locX, this.locY + (double) this.getHeadHeight(), this.locZ);
++
++            ((WorldServer) this.world).sendParticles(this instanceof EntityPlayer ? (EntityPlayer) this : null, EnumParticle.ITEM_CRACK, false, vec3d1.a, vec3d1.b, vec3d1.c, 5, vec3d.a, vec3d.b + 0.05D, vec3d.c, 0.08D, new int[] { Item.getId(itemstack.getItem()), 0 });
++        } else {
++            b(itemstack);
++        }
++    }
++    // PandaSpigot end
++
+     public void b(ItemStack itemstack) {
+         this.makeSound("random.break", 0.8F, 0.8F + this.world.random.nextFloat() * 0.4F);
+ 
++        /* No need to do client-side calculations, addParticle is nullsub on server-side - PandaSpigot
+         for (int i = 0; i < 5; ++i) {
+             Vec3D vec3d = new Vec3D(((double) this.random.nextFloat() - 0.5D) * 0.1D, Math.random() * 0.1D + 0.1D, 0.0D);
+ 
+@@ -848,6 +870,7 @@ public abstract class EntityLiving extends Entity {
+             vec3d1 = vec3d1.add(this.locX, this.locY + (double) this.getHeadHeight(), this.locZ);
+             this.world.addParticle(EnumParticle.ITEM_CRACK, vec3d1.a, vec3d1.b, vec3d1.c, vec3d.a, vec3d.b + 0.05D, vec3d.c, new int[] { Item.getId(itemstack.getItem())});
+         }
++        PandaSpigot end */
+ 
+     }
+ 
+diff --git a/src/main/java/net/minecraft/server/ItemStack.java b/src/main/java/net/minecraft/server/ItemStack.java
+index 015ce283495dc57c8a9dc018ad36a04ebfd637d0..e2cac57120721aaf7adaa3aed236a5226376d9b0 100644
+--- a/src/main/java/net/minecraft/server/ItemStack.java
++++ b/src/main/java/net/minecraft/server/ItemStack.java
+@@ -392,8 +392,9 @@ public final class ItemStack {
+     public void damage(int i, EntityLiving entityliving) {
+         if (!(entityliving instanceof EntityHuman) || !((EntityHuman) entityliving).abilities.canInstantlyBuild) {
+             if (this.e()) {
++                boolean hasUnbreaking = (EnchantmentManager.getEnchantmentLevel(Enchantment.DURABILITY.id, this) > 0); // PandaSpigot - Unbreaking Break Effect Fix
+                 if (this.isDamaged(i, entityliving.bc(), entityliving)) { // Spigot
+-                    entityliving.b(this);
++                    entityliving.renderBrokenItemStack(this, hasUnbreaking); // PandaSpigot - Unbreaking Break Effect Fix
+                     --this.count;
+                     if (entityliving instanceof EntityHuman) {
+                         EntityHuman entityhuman = (EntityHuman) entityliving;
+@@ -417,6 +418,11 @@ public final class ItemStack {
+                     this.damage = 0;
+                 }
+ 
++                // PandaSpigot start - Unbreaking Break Effect Fix
++                if (hasUnbreaking && entityliving instanceof EntityPlayer && this.damage + i > this.j()) {
++                    ((EntityPlayer) entityliving).getBukkitEntity().updateInventory(); // TODO: is it possible to not update entire inventory?
++                }
++                // PandaSpigot end
+             }
+         }
+     }


MasterDash5:
Doesn't the client and server playing the same sound mean the sound will play twice as loud?

anzz1:
Oh right and addParticle leads to a nullsub, there is bunch of particle calculation code all over the codebase that ultimately amounts to nothing. Deleting all that would yield possibly great performance improvements.
Afaik the only particles that are server-side on 1.8.8 are dust clouds when entity falls to ground, dust clouds when breaking an armor stand, and fishing hook's bubbles and waves.

And when it comes to the sounds, later on the entire client-side world sound system was scrapped, and there is no longer a command to play a sound to everyone else except the target player. While the idea of client-side sound system is good on paper, have sounds play in timely manner even in bad network conditions, and reduce amount of packets needed, it probably was source of lot of bugs and mojang decided to not bother with it anymore. In 1.11 all the world sounds are played from server-side.

In 1.11 also this unbreaking issue has been fixed in a brute-force manner, most of the items breaking logic is moved server-side. However from what I can tell they introduced new bug of the particle spawning code never getting called on client so the particles wouldn't show at all. That's what I read from the code, can't be arsed to actually test if its correct.

MasterDash5:
We could spoof the item to remove the Unbreaking enchantment for the client to ensure it doesn't break when we don't want it to. But this is would definitely need to be behind a config option.

anzz1:
> Doesn't the client and server playing the same sound mean the sound will play twice as loud?

No, for the sound waves to constructively interfere like that both sounds would need to be played at *exactly* the same time which is not possible due to natural network delay being non-zero. And in addition, the pitch is randomized which also changes the waveform so volume doubling couldn't have happened for that reason either. The two sounds overlap but not exactly. It actually sounds kinda nice, so it worked out well.

anzz1:
> We could spoof the item to remove the Unbreaking enchantment for the client to ensure it doesn't break when we don't want it to. But this is would definitely need to be behind a config option.

Interesting idea, but wouldn't it break display of enchantments such as this:

ench

And also tooltips when hovering over items in inventory but I guess those can be faked with custom tooltips that lot of servers use, inventories and items for menus when you hover over the tooltip can show whatever.

MasterDash5:
The enchantment display on the screenshot is something from a mod, but it would still break it. Considering we can still spoof the enchantment glint and add to the item lore, it might not be much of a downside, especially if its just done on the last durability.

I do think having a config option for this would be good even without spoofing the item, since there isn't a perfect solution for this.

anzz1:
To be honest that sounds like something that could be done with a plugin instead of putting it into the server core, as it's quite hacky.

I mean if it were possible to start the entire codebase from scratch today, it'd make sense to have the server core just as clean and vanilla with strictly bug fixes and optimizations, and then plenty of APIs that could add additional functionality in a modular fashion. An 'official' repository for the plugin modules could then be maintained.
Unfortunately that's not how it all happened and instead we got an extremely messy codebase that is quite hard to work with, due to the history going all the way back to bukkit and random stuff being added into the codebase for years. Still adding more hacks and mess doesn't seem like a prudent thing to do, if something can't be done with a plugin then just add more hooks more events that expose more of the server so it can.

anzz1:
~~Converted to draft as I'm gonna clean this up a bit.~~

Cleanup done.

uRyanxD:
> Fixes https://github.com/hpfxd/PandaSpigot/issues/190

This PR fixes the problem reported by masterdash in the comments, not the one in the issue itself

anzz1:
> > Fixes #190
> 
> This PR fixes the problem reported by masterdash in the comments, not the one in the issue itself

Hm? They are talking about the same thing?

uRyanxD:
> > > Fixes #190
> > 
> > 
> > This PR fixes the problem reported by masterdash in the comments, not the one in the issue itself
> 
> Hm? They are talking about the same thing?

It's not the same issue; Dash only mentioned it because it's a similar or related bug. The main issue happens even with items that don't have the Unbreakable tag—though it doesn't seem to affect every item—I left a list of the ones I was able to reproduce here:

https://github.com/hpfxd/PandaSpigot/issues/190#issuecomment-3508698004

uRyanxD:
> > > > Fixes #190
> > > 
> > > 
> > > This PR fixes the problem reported by masterdash in the comments, not the one in the issue itself
> > 
> > 
> > Hm? They are talking about the same thing?
> 
> It's not the same issue; Dash only mentioned it because it's a similar or related bug. The main issue happens even with items that don't have the Unbreakable tag—though it doesn't seem to affect every item—I left a list of the ones I was able to reproduce here:
> 
> [#190 (comment)](https://github.com/hpfxd/PandaSpigot/issues/190#issuecomment-3508698004)

This is the issue that the user who created the issue was referring to: https://bugs.mojang.com/browse/MC/issues/MC-2518

I don't see any problem with fixing the one Dash mentioned as well; I'm just saying that the PR fixes the issue Dash pointed out, not the one originally reported by the user

anzz1:
Ohhhh right, hoes can't even have unbreaking. and MC-2518 mentions multiple different items. Hm, I wonder what are the other conditions than unbreaking where the server and client desync on item break. And what other things could create such situation for armor than unbreaking?
I'm just guessing now here but maybe there is some packet ordering problem? As in the whichever action is supposed to break the item comes after an inventory update packet that says the item no longer exist anymore?
Looking into later versions for clues unfortunately won't help here as in 1.11.2 where the bug had been fixed, they just moved to always broadcasting the item break sound from server-side. In fact they seemed to have moved most of all the sounds to server-side apparently they didn't want to keep fixing desync sound bugs anymore

uRyanxD:
> Ohhhh right, hoes can't even have unbreaking. and MC-2518 mentions multiple different items. Hm, I wonder what are the other conditions than unbreaking where the server and client desync on item break. And what other things could create such situation for armor than unbreaking? I'm just guessing now here but maybe there is some packet ordering problem? As in the whichever action is supposed to break the item comes after an inventory update packet that says the item no longer exist anymore? Looking into later versions for clues unfortunately won't help here as in 1.11.2 where the bug had been fixed, they just moved to always broadcasting the item break sound from server-side. In fact they seemed to have moved most of all the sounds to server-side apparently they didn't want to keep fixing desync sound bugs anymore

I don't know, but MC-2518 was fixed in 1.14, not 1.11

uRyanxD:
My best guess is as follows:

```
The item calls ItemStack#takeDamageAndBreak
-> which calls LivingEntity#onBrokenItem
-> which calls Entity#playSound and sends particles
-> Entity#playSound calls world#playSound
-> which calls WorldEventListener#playSound
-> which calls ServerWorldEventListener#playSound
-> which calls PlayerManager.sendPacket and IGNORES THE SOUND SOURCE, which matches what was stated in the bug report comments (other players hear it, but the player themselves does not).
```

My analysis took the Minecraft 1.13 source code into account, but I didn't consider the fact that the particles were also bugged.

Furthermore, Mojang's fix in 1.14 mainly involved moving this to the client side.

Edit: nvm, it sends null as source
EDIT 2: No, I am not mistaken; the player overrides the method and passes itself as the source

anzz1:
Don't you mean mojang moving it to the *server* side, as thats what has happened in 1.11.2 (at the latest). 

And yes the way playsound for EntityPlayer works is that it calls the playSoundToNearExcept, deliberately skipping sending the sound packet to the player that triggered it because its supposed to be played client-side. If the player doesn't hear it themselves its precisely a desync, the server doing what it should skipping the player, but the client side player not playing the sound that it should.
So what needs to be looked for is the reason why the client doesn't reach its own playSound that its supposed to.
And probably due to being exhausted with these type of desync bugs mojang just moved it all to server side when sure they lose the original idea, that is sounds playing with zero delay for the player itself, plus not requiring a packet to play sound, but at same time they fix all these desync bugs.

Oh right and at least in 1.11.2 they didn't actually fix the particles so them moving the whole renderbrokenitemstack to server-side without moving the particles makes the particles never work. Probably they fixed that in some later version, but havent looked through any diffs beyond 1.11.2.

uRyanxD:
> Don't you mean mojang moving it to the _server_ side, as thats what has happened in 1.11.2 (at the latest).
> 
> And yes the way playsound for EntityPlayer works is that it calls the playSoundToNearExcept, deliberately skipping sending the sound packet to the player that triggered it because its supposed to be played client-side. If the player doesn't hear it themselves its precisely a desync, the server doing what it should skipping the player, but the client side player not playing the sound that it should. So what needs to be looked for is the reason why the client doesn't reach its own playSound that its supposed to. And probably due to being exhausted with these type of desync bugs mojang just moved it all to server side when sure they lose the original idea, that is sounds playing with zero delay for the player itself, plus not requiring a packet to play sound, but at same time they fix all these desync bugs.
> 
> Oh right and at least in 1.11.2 they didn't actually fix the particles so them moving the whole renderbrokenitemstack to server-side without moving the particles makes the particles never work. Probably they fixed that in some later version, but havent looked through any diffs beyond 1.11.2.

What I meant was that MC-1518 was only fixed in 1.14, and Mojang moved `onBrokenItem` (Feather mapping) to the client side in that version.

https://prnt.sc/zLsLbBuy6BsP

It seems the server simply triggers an event intended to notify the client, or something of that sort; I didn't delve too deeply into it.


(Sorry if this is hard to understand; English isn't my native language.)

anzz1:
> What I meant was that MC-1518 was only fixed in 1.14, and Mojang moved onBrokenItem (Feather mapping) to the client side in that version.

Ah, that would mean then that they moved the onBrokenItem *back* to being client-side, same as it was in 1.8, while in 1.11 it was all server-side :sweat_smile: As that is how they fixed the [MC-1040](https://bugs.mojang.com/browse/MC/issues/MC-1040) in [16w43a](https://minecraft.wiki/w/Java_Edition_16w43a) (snapshot of 1.11). So apparently they half-fixed the issue in 1.11, fixing sounds but breaking particles by moving the sounds but not the particles to server-side, then in 1.14 they moved the item breaking back to client-side :laughing: 

Your command of English is just fine, its not my native either, its rather these back-and-forth changes that Mojang makes that throws us all off :smile: 

The site [mcsrc.dev](https://mcsrc.dev/) is fantastic for diffing more modern versions, but unfortunately the oldest version it has is 1.14.4. I've been using the [modcoderpack.com](http://www.modcoderpack.com/) sources for diffing those earlier versions, but it'd be nice if I could figure out how to diff all of the versions including the older ones

anzz1:
I converted this to Draft as discussed in https://github.com/hpfxd/PandaSpigot/pull/436#issuecomment-4931258932 , I agree that we should join these issues and fix the breaking sounds once and for all, not just for unbreaking, in one go.

So to reiterate from the other issue, the brute-force method of broadcasting the item breaking effects could be done for all items, not just unbreaking, that it would always broadcast the item breaking sound and effect to all players including the one who triggered it, but this causes the sound play twice and particles duplicate if the player is *not* desynced. This is an okay fix, but better would be to investigate the root cause if its possible to make sure the client stays synced in the non-unbreaking equipment breaks. 
And as @uRyanxD caught in the other issue, the method used here to play the sound should be changed that it still plays the `PlayerSoundEvent` and not `EntitySoundEvent`.

uRyanxD:
Another amazing bug :)

https://bugs.mojang.com/browse/MC/issues/MC-10369

anzz1:
@uRyanxD 
Oh yeah that is the bug I mentioned here: https://github.com/hpfxd/PandaSpigot/pull/412#issuecomment-4762243831
That when they fixed the sound issue somewhere between 1.9-1.11 by moving all the client side sounds to server side, they didnt follow the code to end and made it so the whole sound **and** particle code was also guarded by the !isRemote, but they only changed the sound part so to server side but kept the particle code as-is, which is supposed to be called from client side, and is just a nullsub on server side, stopping the particles from ever showing.

It's precisely why its correctly fixed here with the `renderBrokenItemStack` with the broadcast flag, so when its played from server side it actually send the particle spawn packet.
Anyway though like you mentioned on the other PR this need to be bolted into the playersoundevent, and also you had the broken item issue trigger on non-unbreaking items too, so the whole unbreaking logic could be just axed and always used the broadcast version of renderbrokenitemstack when players item breaks.

I havent been able to replicate that yet actually, the issue on items without unbreaking, how do you trigger exactly the sound not playing/particles not showing in those?

uRyanxD:
> @uRyanxD Oh yeah that is the bug I mentioned here: [#412 (comment)](https://github.com/hpfxd/PandaSpigot/pull/412#issuecomment-4762243831) That when they fixed the sound issue somewhere between 1.9-1.11 by moving all the client side sounds to server side, they didnt follow the code to end and made it so the whole sound **and** particle code was also guarded by the !isRemote, but they only changed the sound part so to server side but kept the particle code as-is, which is supposed to be called from client side, and is just a nullsub on server side, stopping the particles from ever showing.
> 
> It's precisely why its correctly fixed here with the `renderBrokenItemStack` with the broadcast flag, so when its played from server side it actually send the particle spawn packet. Anyway though like you mentioned on the other PR this need to be bolted into the playersoundevent, and also you had the broken item issue trigger on non-unbreaking items too, so the whole unbreaking logic could be just axed and always used the broadcast version of renderbrokenitemstack when players item breaks.
> 
> I havent been able to replicate that yet actually, the issue on items without unbreaking, how do you trigger exactly the sound not playing/particles not showing in those?

Contrary to what is stated in issue https://bugs.mojang.com/browse/MC/issues/MC-2518, I was unable to reproduce the problem with all the items listed when I tested it; the ones that did trigger it were:

Hoes, Swords, Fishing Rods, and Armor

The easiest way is to use `/give` to get a hoe with 1 durability point remaining and till the soil.

The "magic" solution for many of these issues in version 1.14 was to move the handling logic to the client side, with the server simply sending a "WorldEvent" packet containing the specific event ID (e.g., for particles or sounds).

anzz1:
Thanks, I'll try to reproduce it with those instructions to try to understand what its about.

> Hoes, Swords, Fishing Rods, and Armor

This makes it even more strange, like why those but not Pickaxes, Axes, Bows, Shears, Flint and Steel, or Carrot on a Stick. It'd make much more sense if it applied to all items instead of just some of them.

PR #416, opened 2026-06-28
fix configurations

anzz1:
a double \n\n cuts the description off.

### DIFF
--- a/patches/server/0016-Configurable-time-update-frequency.patch
+++ b/patches/server/0016-Configurable-time-update-frequency.patch
@@ -8,12 +8,11 @@ diff --git a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.j
 index 3dd3d0702c17541083d0d365ca46e93602105bce..b5f16e8cab9e43ad7597b2ea5bc921399b37872b 100644
 --- a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
 +++ b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
-@@ -6,6 +6,13 @@ import org.spongepowered.configurate.objectmapping.meta.Comment;
+@@ -6,6 +6,12 @@ import org.spongepowered.configurate.objectmapping.meta.Comment;
  @ConfigSerializable
  @SuppressWarnings({"FieldCanBeLocal", "FieldMayBeFinal"})
  public class PandaSpigotWorldConfig {
 +    @Comment("How many ticks in between sending time updates to players?\n" +
-+            "\n" +
 +            "The vanilla option is 20 (every second), but PandaSpigot sets the default\n" +
 +            "to 100 (every 5 seconds). You would probably be fine setting this even\n" +
 +            "higher, unless you're constantly changing the time, or the server is lagging.")
--- a/patches/server/0017-Smooth-teleportation.patch
+++ b/patches/server/0017-Smooth-teleportation.patch
@@ -11,14 +11,13 @@ diff --git a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.j
 index b5f16e8cab9e43ad7597b2ea5bc921399b37872b..bde014f003bccca916dbc74fc27436bf4b193fdf 100644
 --- a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
 +++ b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
-@@ -13,6 +13,15 @@ public class PandaSpigotWorldConfig {
+@@ -12,6 +12,14 @@ public class PandaSpigotWorldConfig {
              "higher, unless you're constantly changing the time, or the server is lagging.")
      public int timeUpdateFrequency = 100;
      
 +    @Comment("This option makes it so that when players are teleported to a location\n" +
 +            "with the same rotation they currently have, the server will send a special\n" +
 +            "packet indicating that the client should not update it's rotation at all.\n" +
-+            "\n" +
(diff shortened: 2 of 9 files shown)


uRyanxD:
> a double \n\n cuts the description off.

Are you sure? It works fine for me.

https://prnt.sc/Qey8Uo69WOF-

The AI ​​and time-related settings also use two \n and seem okay to me.

anzz1:
Yeah, cut off for me, only the last section after last \n\n is printed to the config
```
    # This option controls whether or not there is a chance for arrow crits to deal extra damage.
    # By default, this is true (vanilla behaviour)
    randomArrowDamage: true
    # Default: false
    useSpigotChunkTickingSelection: false
```

Java 17 (the default from package manager) on Debian 12 (Bookworm) if it makes any difference 
```
openjdk 17.0.19 2026-04-21
OpenJDK Runtime Environment (build 17.0.19+10-1-deb12u2-Debian)
OpenJDK 64-Bit Server VM (build 17.0.19+10-1-deb12u2-Debian, mixed mode, sharing)
```

uRyanxD:
> Yeah, cut off for me, only the last section after last \n\n is printed to the config
> 
> ```
>     # This option controls whether or not there is a chance for arrow crits to deal extra damage.
>     # By default, this is true (vanilla behaviour)
>     randomArrowDamage: true
>     # Default: false
>     useSpigotChunkTickingSelection: false
> ```
> 
> Java 17 (the default from package manager) on Debian 12 (Bookworm) if it makes any difference
> 
> ```
> openjdk 17.0.19 2026-04-21
> OpenJDK Runtime Environment (build 17.0.19+10-1-deb12u2-Debian)
> OpenJDK 64-Bit Server VM (build 17.0.19+10-1-deb12u2-Debian, mixed mode, sharing)
> ```

Does this happen with the AI ​​and time settings in the Panda config as well? If not, it's probably a bug in how Configurate adds new sections to existing configurations, or something like that

anzz1:
> Does this happen with the AI ​​and time settings in the Panda config as well? If not, it's probably a bug in how Configurate adds new sections to existing configurations, or something like that

Oh yeah, it does indeed.
```
  default:
    # The vanilla option is 20 (every second), but PandaSpigot sets the default
    # to 100 (every 5 seconds). You would probably be fine setting this even
    # higher, unless you're constantly changing the time, or the server is lagging.
    timeUpdateFrequency: 20
    # For example, normally constantly teleporting a player to their own location
    # will make moving their head very difficult, especially for players with higher latency.
    # With this option enabled, they will be able to move their head just like normal.
    smoothTeleportation: true
    # When enabled, this option disables reading and writing player data such as:
    # - Position
    # - Inventory
    # - Enderchest
    disablePlayerData: false
    # When enabled, this option will disable saving world chunks.
    disableChunkSaving: false
    # This has the same effect as Spigot's "nerf-spawner-mobs" option, but applies to all entities.
    disableEntityAi: false
```
So this patch should be extended to not just fix the ACTS one but the others too.
Just a minute.

uRyanxD:
Never mind what I wrote and deleted; I didn't see what was right in front of me. I find it a bit strange that this doesn't happen on Windows, and I don't have WSL set up to test whether the issue occurs on another Linux distribution or just this specific one

anzz1:
It could be that whatever generates the configuration treates \n\n as EOF marker, but windows version presumably uses \r\n\r\n instead. There is always bunch of weirdness related to line breaks especially when dealing with "multiplatform" high-level language stuff because they attempt to hide these platform specific newline differences and then stuff breaks when some lib function does/doesn't consider \r\n equal to \n, or eats/adds the \r when you were actually working with raw bytes instead of any text, and so on.

Anyway I updated the commit to update the other patches that also had \n\n in them, and the subsequent patches that touch `PandaSpigotWorldConfig.java` so they apply cleanly.

This one is probably wise to merge first so rest of the PRs that touch the config can be properly rebased on this.

anzz1:
> Never mind what I wrote and deleted; I didn't see what was right in front of me.

Exactly why this PR only fixed the ACTS one and not the others first because I didn't see them either :D

uRyanxD:
There's another case in that file, lol
```random Arrow Trajectory```
And also in the non-per-world config
```worlds node comment```
```trackPluginScoreboards```

**Comment by anzz1** [comment]
Wait, where
`0030-Configurable-arrow-trajectory.patch` looks fine?
`0038-Lazily-track-plugin-scoreboards-by-default.patch` looks good too

`worlds node comment` i can't find this one my grep turns up empty

edit: Ah you mean this one
`CommentedConfigurationNode worldsNode = root.node("worlds")` in `PandaSpigotConfig.java`
that one seems to work since its not a config option but rather a heading for the "worlds" section
```
# The worlds section is for settings which can be configured per-world.
# 
# Any settings in the "default" world will provide default values for
# other worlds which don't explicitly specify settings.
# 
# To specify settings for a specific world, just add a new section with the world's name.
```
Although to be consistent and maybe avoid some migration or etc. problems it could also have its line breaks removed

uRyanxD:
> Wait, where `0030-Configurable-arrow-trajectory.patch` looks fine? `0038-Lazily-track-plugin-scoreboards-by-default.patch` looks good too
> 
> `worlds node comment` i can't find this one my grep turns up empty
> 
> edit: Ah you mean this one `CommentedConfigurationNode worldsNode = root.node("worlds")` in `PandaSpigotConfig.java` that one seems to work since its not a config option but rather a heading for the "worlds" section
> 
> ```
> # The worlds section is for settings which can be configured per-world.
> # 
> # Any settings in the "default" world will provide default values for
> # other worlds which don't explicitly specify settings.
> # 
> # To specify settings for a specific world, just add a new section with the world's name.
> ```
> 
> Although to be consistent and maybe avoid some migration or etc. problems it could also have its line breaks removed

Forget the first ones; I opened the wrong Panda folder (that one actually had double `\n` characters in them, because I was running tests) 😂

uRyanxD (approved):

PR #417, opened 2026-06-28
Fix MC-233911 Wakeup Time Sync

anzz1:
Sends a time sync packet immediately after waking up. This should make players wake up to daylight as long as they aren't lagging too much.

Fixes the issue that it takes up to `timeUpdateFrequency` ticks in pandaspigot.yml + network delay for players to have their time refreshed after waking up, so they'd wake up to it still being night that suddenly turns into a day.
As a sidenote I don't really understand what was the reasoning for the `timeUpdateFrequency` to be configurable in the first place, its not as if sending a small packet to all players once per second had any performance effect.

With the update frequency default of 5 seconds this is most helpful as up to 5 seconds + network delay players spend still in night after waking up without the patch. But even with vanilla value of 1 second, this still helps as now you should be waking up to daylight from bed without seeing any night at all as long as you don't have a lot of network delay.

### DIFF
--- a/patches/server/0133-Fix-MC-233911-Wakeup-Time-Sync.patch
+++ b/patches/server/0133-Fix-MC-233911-Wakeup-Time-Sync.patch
@@ -1,39 +1,48 @@
 From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
 From: anzz1 <anzz1@live.com>
-Date: Sun, 28 Jun 2026 17:18:47 +0000
+Date: Mon, 27 Jul 2026 08:03:39 +0000
 Subject: [PATCH] Fix MC-233911 Wakeup Time Sync
 
-Sends a time sync packet immediately after waking up.
-This should make players wake up to daylight as long as they aren't
-lagging too much.
+Sends a time sync packet immediately after waking up from sleep.
 
+diff --git a/src/main/java/net/minecraft/server/MinecraftServer.java b/src/main/java/net/minecraft/server/MinecraftServer.java
+index ae0259e025d1121adb5d09cda05e84fe711d8a35..d5993c69ffcfba7334d4a614b056b665978e6622 100644
+--- a/src/main/java/net/minecraft/server/MinecraftServer.java
++++ b/src/main/java/net/minecraft/server/MinecraftServer.java
+@@ -857,7 +857,7 @@ public abstract class MinecraftServer implements Runnable, ICommandListener, IAs
+             long worldTime = world.getTime();
+             final PacketPlayOutUpdateTime worldPacket = new PacketPlayOutUpdateTime(worldTime, dayTime, doDaylight);
+             for (EntityHuman entityhuman : world.players) {
+-                if (!(entityhuman instanceof EntityPlayer) || (this.ticks + entityhuman.getId()) % world.pandaSpigotConfig.timeUpdateFrequency != 0) {
++                if (!(entityhuman instanceof EntityPlayer) || (!world.timeNeedsSync && (this.ticks + entityhuman.getId()) % world.pandaSpigotConfig.timeUpdateFrequency != 0)) { // PandaSpigot - Wakeup Time Sync
+                     continue;
+                 }
+                 EntityPlayer entityplayer = (EntityPlayer) entityhuman;
+@@ -866,6 +866,7 @@ public abstract class MinecraftServer implements Runnable, ICommandListener, IAs
+                         new PacketPlayOutUpdateTime(worldTime, playerTime, doDaylight);
+                 entityplayer.playerConnection.sendPacket(packet); // Add support for per player time
+             }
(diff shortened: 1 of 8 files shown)


overkidding:
Tho this means that it could send 2 packets to the same player if it happens in the same tick as the timeUpdateFrequency or am i wrong?

anzz1:
Yes that is possible since there is no guard against whether a time update packet has already been queued this tick.
But considering it is such a small packet, and on the client-side all that is processed is setting the worldTime variable in worldinfo when receiving the packet, setting up such guard is unnecessary.
What I'm saying this packet is very inexpensive hence I don't think the configurable time updates patch and setting default to 5 seconds up from 1 was ever necessary, nor does it matter if the client receives the same packet twice in a tick.

And overall, there is no packet de-duplication mechanism in vanilla, in this case again as the packet is so small and inconsequential setting up such system for this specifically wouldn't make sense and would probably lose performance compared to just sending the packet, but overall such mechanism could *maybe* be useful however it could get complex to parse which are types of packets that should be de-duplicated and which not.

uRyanxD:
Just for reference, this is a very old bug: https://bugs.mojang.com/browse/MC/issues/MC-233911; it was recently fixed in version 26.1

anzz1:
Added the MC-233911 identifier to the patch

anzz1:
Actually now that I think of it, even if the players do have delay it should still work anyway just fine, considering that the wakeup packet and time sync packet get sent in the same tick and should arrive at the same time even with delay.

Although I did wonder whether the whole `if (this.everyoneDeeplySleeping())` block could be moved to the same block with the `if (this.getGameRules().getBoolean("doDaylightCycle"))` block reducing code and making sure that the time change packet arrives first before the wake up packet, but ultimately didn't move it as it seems so obvious at a glance that it should be in the same block that maybe there is some reason the logic is done in two blocks like this, before and after the mob spawner (and chunk unloader and setskylightsubstracted)? or is it just shoddy code? idk.

patches/server/0133-Fix-MC-233911-Wakeup-Time-Sync.patch
```diff
@@ -0,0 +1,39 @@
...
+             }
+ 
+             this.e();
+@@ -243,6 +245,11 @@ public class WorldServer extends World implements IAsyncTaskHandler {
+         this.worldData.setTime(this.worldData.getTime() + 1L);
+         if (this.getGameRules().getBoolean("doDaylightCycle")) {
+             this.worldData.setDayTime(this.worldData.getDayTime() + 1L);
++            // PandaSpigot start - Wakeup Time Sync
++            if (timeNeedsSync) {
++                this.server.getPlayerList().a(new PacketPlayOutUpdateTime(this.worldData.getTime(), this.worldData.getDayTime(), true), this.worldProvider.getDimension());
```
uRyanxD: LGTM; the only thing I think needs changing is that,
unlike the patch at https://github.com/hpfxd/PandaSpigot/blob/master/patches/server/0015-Optimize-World-Time-Updates.patch#L10
here we aren't dealing with the (stupid) Bukkit `setPlayerTime(boolean relative)` API.

anzz1: sorry i don't quite understand, changing to what? did you mean to use the submit changes function? i don't see anything changed

uRyanxD: > sorry i don't quite understand, changing to what? did you mean to use the submit changes function? i don't see anything changed

What I meant is that we should handle the `player#setPlayerTime` API here, just as we do in https://github.com/hpfxd/PandaSpigot/blob/master/patches/server/0015-Optimize-World-Time-Updates.patch; otherwise, we'll always be sending the world time instead of the time defined by the API. This API is quite strange, but I believe this is the expected behavior.

anzz1: ahhhh got it, i'll fix that

anzz1: Alright I changed it, but I'm not sure if the time sync packet still get sent on the same tick, or on the next one?

uRyanxD: I believe it's on the same tick, but I'm not 100% sure.

anzz1: alright i'm gonna test it to make sure

anzz1: alright it didn't send on the same tick, but i fixed it now

anzz1: it needed a refactor to make it work sensibly with the custom player time api but now its ok i think

uRyanxD (approved):

PR #418, opened 2026-06-28
docs: document Windows Git Bash workflow

wilywork:
## Summary
- Adds README instructions for building PandaSpigot on Windows using Git Bash.
- Documents when to run `./panda setup` and `./panda jar`.
- Adds the patch workflow for `PandaSpigot-Server` and `PandaSpigot-API`.

## Testing
- Not run; documentation-only change.


patches/server/0131-Add-growable-plant-abstraction.patch
uRyanxD, line 1: Unrelated patch?

README.md
uRyanxD, line 1: I'm not 100% sure about this, but it seems unnecessary to me; for example, the section on how to create patches is already covered by Paper's CONTRIBUTING.md guide, which is linked in the README.md.

MasterDash5:
This should probably be modified to explain that these steps work with other operating systems, as long as they have bash.

wilywork:
Thanks for the feedback.

I agree that the detailed patch creation steps are already covered by Paper's CONTRIBUTING.md, so I trimmed this down and kept it focused on PandaSpigot-specific context.

The intent here is mainly to clarify that `panda` is a Bash script and that the documented commands should work on any platform as long as they are run from a Bash-compatible shell.

I also kept a short note about the generated `PandaSpigot-API` and `PandaSpigot-Server` work trees because that part can be easy to misunderstand when first contributing to this patch-based workflow.

uRyanxD (approved):
LGTM, thank you

PR #419, opened 2026-06-29
update to 135 to fix several CVEs

TH3N3WN00B:
this is my first time doing this, please close and provide feedback on how to do it correctly if made wrong :)

### DIFF
--- a/patches/server/0005-Update-to-Netty-4.1.x.patch
+++ b/patches/server/0005-Update-to-Netty-4.1.x.patch
@@ -3,6 +3,335 @@ From: hpfxd <me@hpfxd.com>
 Date: Wed, 27 Oct 2021 06:29:44 -0400
 Subject: [PATCH] Update to Netty 4.1.x
 
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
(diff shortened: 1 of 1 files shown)


uRyanxD:
It looks like you duplicated the patch content within the patch itself. Also, please do not edit the patches manually; use our tool "./panda edit server".

TH3N3WN00B:
I did use ./panda edit server... but didnt used vim used nano... for some reason (most likely im dumb) the list of patches doesnt appear with vim

I just wanted to update netty nothing else, so the patch should be the same

uRyanxD:
> I did use ./panda edit server... but didnt used vim used nano... for some reason (most likely im dumb) the list of patches doesnt appear with vim
> 
> I just wanted to update netty nothing else, so the patch should be the same

So, it looks like Git is broken on your system or something, because it duplicated the patch content within itself instead of replacing the content :p

uRyanxD:
> So, it looks like Git is broken on your system or something, because it duplicated the patch content within itself instead of replacing the content :p

Also, from what I've seen, none of these CVEs seem to affect Minecraft 1.8/PandaSpigot.

uRyanxD:
Closing this PR due to the reasons cited above.

PR #421, opened 2026-06-29
Projectile Fixes

### LINKED ISSUE #420: [MC-51126] Bows not working on boats
Bows don't work properly when riding a boat.

If you have a bow without Infinity, then sometimes when you shoot at high angles it somewhat works but the arrow gets shot weirdly sideways, while if you shoot at low angles the arrow doesn't fly at all, and sometimes hits yourself. There is also weirdness whether the *twang* sound from the bow gets played at all, or if it gets played with a delay. Somehow also shooting at low angle can break the high angle and now that doesn't work either.

Bow with Infinity never shoots the arrow at all. And it seems, even though you dont see it, as if the arrow shoots straight up from your position and then hits you if you charge the bow more. If you charge the bow fully and shoot at some direction and wait a bit, the arrow will hit and kill you.

This is most likely related to [MC-51126](https://bugs.mojang.com/browse/MC/issues/MC-51126). The fix needs to be backported from [20w21a](https://minecraft.wiki/w/Java_Edition_20w21a) if the changelog is correct and that is the version that fixed the bug.

**Steps to reproduce**

Get yourself a bow with Infinity, another one without, and a boat
```
/give @p minecraft:bow 1 0 {ench:[{id:34,lvl:3s},{id:51,lvl:1s}]}
/give @p minecraft:bow
/give @p minecraft:arrow 64
/give @p minecraft:boat
```
Place the boat on water, ride it, shoot bow at different angles. Observe weirdness.

anzz1:
Make projectile not collide with the ridden entity for first few ticks, same as player

Fixes #420 ([MC-51126](https://bugs.mojang.com/browse/MC/issues/MC-51126))

### DIFF
--- a/patches/server/0136-Projectile-Fixes.patch
+++ b/patches/server/0136-Projectile-Fixes.patch
@@ -1,0 +1,59 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Sat, 8 Aug 2026 19:54:03 +0000
+Subject: [PATCH] Projectile Fixes
+
+- Make projectile not collide with the ridden entity for first few ticks
+
+diff --git a/src/main/java/net/minecraft/server/EntityArrow.java b/src/main/java/net/minecraft/server/EntityArrow.java
+index ecde0d9ec034024355c09b249694ce118b3b10bc..1cfea4ba17c64ff6534d6898408ed9ddd3eb6fd6 100644
+--- a/src/main/java/net/minecraft/server/EntityArrow.java
++++ b/src/main/java/net/minecraft/server/EntityArrow.java
+@@ -191,7 +191,7 @@ public class EntityArrow extends Entity implements IProjectile {
+             for (j = 0; j < list.size(); ++j) {
+                 Entity entity1 = (Entity) list.get(j);
+ 
+-                if (entity1.ad() && (entity1 != this.shooter || this.as >= 5)) {
++                if (entity1.ad() && ((entity1 != this.shooter && entity1.passenger != this.shooter) || this.as >= 5)) { // PandaSpigot - Projectile Fixes
+                     f1 = 0.3F;
+                     AxisAlignedBB axisalignedbb1 = entity1.getBoundingBox().grow((double) f1, (double) f1, (double) f1);
+                     MovingObjectPosition movingobjectposition1 = axisalignedbb1.a(vec3d, vec3d1);
+diff --git a/src/main/java/net/minecraft/server/EntityFireball.java b/src/main/java/net/minecraft/server/EntityFireball.java
+index 2524e2b8d8e03e4c97464eb74a964b595584f8a7..0a3a5d9370a56056abdbd33e7bf5993b81c5d9cc 100644
+--- a/src/main/java/net/minecraft/server/EntityFireball.java
++++ b/src/main/java/net/minecraft/server/EntityFireball.java
+@@ -106,7 +106,7 @@ public abstract class EntityFireball extends Entity {
+             for (int i = 0; i < list.size(); ++i) {
+                 Entity entity1 = (Entity) list.get(i);
+ 
+-                if (entity1.ad() && (!entity1.k(this.shooter) || this.as >= 25)) {
++                if (entity1.ad() && ((!entity1.k(this.shooter) && entity1.passenger != this.shooter) || this.as >= 25)) { // PandaSpigot - Projectile Fixes
+                     float f = 0.3F;
+                     AxisAlignedBB axisalignedbb = entity1.getBoundingBox().grow((double) f, (double) f, (double) f);
+                     MovingObjectPosition movingobjectposition1 = axisalignedbb.a(vec3d, vec3d1);
+diff --git a/src/main/java/net/minecraft/server/EntityFishingHook.java b/src/main/java/net/minecraft/server/EntityFishingHook.java
+index 6eca198f9f026c82b8b31d446c3953dabdc956cf..6630761c1fcbd6c1588252cb2df28123f7577295 100644
+--- a/src/main/java/net/minecraft/server/EntityFishingHook.java
++++ b/src/main/java/net/minecraft/server/EntityFishingHook.java
+@@ -168,7 +168,7 @@ public class EntityFishingHook extends Entity {
+             for (int i = 0; i < list.size(); ++i) {
+                 Entity entity1 = (Entity) list.get(i);
+ 
+-                if (entity1.ad() && (entity1 != this.owner || this.au >= 5)) {
++                if (entity1.ad() && ((entity1 != this.owner && entity1.passenger != this.owner) || this.au >= 5)) { // PandaSpigot - Projectile Fixes
+                     float f = 0.3F;
+                     AxisAlignedBB axisalignedbb = entity1.getBoundingBox().grow((double) f, (double) f, (double) f);
+                     MovingObjectPosition movingobjectposition1 = axisalignedbb.a(vec3d, vec3d1);
+diff --git a/src/main/java/net/minecraft/server/EntityProjectile.java b/src/main/java/net/minecraft/server/EntityProjectile.java
+index fba0c18c6e1128d2f627c176531e4afe54bcd7df..fa69eafadea380aff012723f0c42ff345ec6b51b 100644
+--- a/src/main/java/net/minecraft/server/EntityProjectile.java
++++ b/src/main/java/net/minecraft/server/EntityProjectile.java
+@@ -126,7 +126,7 @@ public abstract class EntityProjectile extends Entity implements IProjectile {
+             for (int i = 0; i < list.size(); ++i) {
+                 Entity entity1 = (Entity) list.get(i);
+ 
+-                if (entity1.ad() && (entity1 != entityliving || this.ar >= 5)) {
++                if (entity1.ad() && ((entity1 != entityliving && entity1.passenger != entityliving) || this.ar >= 5)) { // PandaSpigot - Projectile Fixes
+                     float f = 0.3F;
+                     AxisAlignedBB axisalignedbb = entity1.getBoundingBox().grow((double) f, (double) f, (double) f);
+                     MovingObjectPosition movingobjectposition1 = axisalignedbb.a(vec3d, vec3d1);
--- a/patches/server/0136-Projectile-Fixes.patch
+++ b/patches/server/0136-Projectile-Fixes.patch
@@ -3,7 +3,7 @@ From: anzz1 <anzz1@live.com>
 Date: Sat, 8 Aug 2026 19:54:03 +0000
 Subject: [PATCH] Projectile Fixes
 
-- Make projectile not collide with the ridden entity for first few ticks
+Make projectile not collide with the ridden entity for first few ticks
 
 diff --git a/src/main/java/net/minecraft/server/EntityArrow.java b/src/main/java/net/minecraft/server/EntityArrow.java
 index ecde0d9ec034024355c09b249694ce118b3b10bc..1cfea4ba17c64ff6534d6898408ed9ddd3eb6fd6 100644


anzz1:
This one is not perfect, the client still sometimes eats the arrow if the fix is not applied to client-side too.
I'm not exactly sure how to alleviate this.

anzz1:
Quoting post from the discord here for posterity, talking about the "projectiledesyncfix" in https://github.com/ThatApplePieGuy/PacketPatcher/

> the other patches look good at quick look, but im not so sure about the projectiledesyncfix, if i understand what you're doing correctly your simply eschewing the projectile sync packets altogether and solely relying the client to be correct?
> when i was testing for the riding projectile sync fix here https://github.com/hpfxd/PandaSpigot/pull/421/
> even without it I noticed that server and client dont always agree on collisions even in a low ping (50ms, stable, no packet loss) environment, i dont think the clientside position of vehicles is always 100% correct on 1.8.8, which cause issue like the client thinking the arrow collided with the vehicle and eating it, yet the hit *twang* sound comes from server so the player still hears the invisible arrow hitting
> this is what players are talking about in the MC-51126
> https://bugs.mojang.com/browse/MC/issues/MC-51126
> as the bug is two-fold, first both the server and client dont have the same "first few ticks" collision immunity for your vehicle as it does for the player that it does for the player so the projectile immediately hits ur vehicle on spawn, which this PR fixes (for the server), but the other part of the bug is that client and server dont agree on the collision so apparently sometimes the vehicle on the server side, with or without this fix, is not on the way while the client think it is, making server shoot an arrow but the client collision eating it and not tracking it
> adding the PR fixes the issue for vehicles server-side, but unless the client-side is patched too sometimes the client still eats the arrow and it becomes invisible but this happens much less often even without client-side patch. so somehow the arrow gets tracked even if the client is not patched
> so instead of trusting the client, could this be done the other way that is always forcing a sync for projectiles and never trusting the client?

anzz1:

overkidding:
> I'm not exactly sure how to alleviate this.

Cant you re-track the projectile after x ticks?

anzz1:
Yeah that was what I was wondering too, could the projectiles simply be re-tracked after 25 ticks for Fireball or 5 ticks for other projectile are passed.

But apparently there is more than that to it since the whole projectile system in MC 1.8.9 is pretty broken, as pointed out in this discord message by the user @ThatApplePieGuy 
> what i wanted to solve was projectiles appearing to jump forward/backward, (easier to see with hitboxes on) i tried updating the resync frequency to be every tick instead of every second but it resulted in projectiles jumping erratically, and that was on localhost without potential ping spikes
the resyncs also can't be trusted, the server ticks projectile physics and then ticks the entitytracker, while the client tick first receives incoming packets and then ticks projectile physics, so every resync packet ends up placing the projectile a tick ahead for the same tick snapshot. the velocity may or may not be a tick ahead depending on whether the projectile type sends velocity updates, and whether the projectile ends up a tick ahead within the initial spawn packet depends on whether you were inside the projectile's initial tracking range or entered it later
lots of annoying subtick jank

This was about his plugin here https://github.com/ThatApplePieGuy/PacketPatcher/ that simply doesn't send the projectile sync packets at all, always relying only on the client, because he was having issues with projectiles jumping around on client-side.
So it seems there is no perfect fix possible without fixing the whole projectile system at both server and client side.

Maybe the needed hooks to the projectile system could be made to provide an API that could then be used by a plugin that sends different projectile packets that would be then received by client side plugin.
I've created this simple library [CapabilitiesLib](https://ftp.taco.cab/games/mc/pandaspigot/plugins/CapabilitiesLib-1.0.jar) which can advertise extended capabilities on connect to clients which can then be used by client-side patches to enable fixes/features that are advertised by server.
Its as simple as importing the lib to a plugin and then using `CapabilitiesLib.addCapability(String capabilityIdentifier, capabilityVersion);` which will then add it to array of capabilities sent to clients on join.

overkidding:
> Yeah that was what I was wondering too, could the projectiles simply be re-tracked after 25 ticks for Fireball or 5 ticks for other projectile are passed.
> 
> But apparently there is more than that to it since the whole projectile system in MC 1.8.9 is pretty broken, as pointed out in this discord message by the user @ThatApplePieGuy 
> > what i wanted to solve was projectiles appearing to jump forward/backward, (easier to see with hitboxes on) i tried updating the resync frequency to be every tick instead of every second but it resulted in projectiles jumping erratically, and that was on localhost without potential ping spikes
> the resyncs also can't be trusted, the server ticks projectile physics and then ticks the entitytracker, while the client tick first receives incoming packets and then ticks projectile physics, so every resync packet ends up placing the projectile a tick ahead for the same tick snapshot. the velocity may or may not be a tick ahead depending on whether the projectile type sends velocity updates, and whether the projectile ends up a tick ahead within the initial spawn packet depends on whether you were inside the projectile's initial tracking range or entered it later
> lots of annoying subtick jank
> 
> This was about his plugin here https://github.com/ThatApplePieGuy/PacketPatcher/ that simply doesn't send the projectile sync packets at all, always relying only on the client, because he was having issues with projectiles jumping around on client-side.
> So it seems there is no perfect fix possible without fixing the whole projectile system at both server and client side.
> 
> Maybe the needed hooks to the projectile system could be made to provide an API that could then be used by a plugin that sends different projectile packets that would be then received by client side plugin.
> I've created this simple library [CapabilitiesLib](https://ftp.taco.cab/games/mc/pandaspigot/plugins/CapabilitiesLib-1.0.jar) which can advertise extended capabilities on connect to clients which can then be used by client-side patches to enable fixes/features that are advertised by server.
> Its as simple as importing the lib to a plugin and then using `CapabilitiesLib.addCapability(String capabilityIdentifier, capabilityVersion);` which will then add it to array of capabilities sent to clients on join.

Or by a little of network performance loss we could track all projectiles every tick instead and try to predict if the arrow got "eaten" and re-sync it.

anzz1:
Yeah I'm open to including more fixes in this patch if someone finds out ways to make the projectiles even better, its part reason why I named the patch "Projectile Fixes" even though it currently only has one fix, that it can be extended with more.

uRyanxD:
Won't it keep happening if the player is a passenger?

Also, couldn't we backport Mojang's solution? It seems more complete to me.

-------------

That’s just an example of their fix—untested in production and likely containing errors or missing parts, since I simply instructed the AI ​​to port it quickly and there are many different mappings involved, not to mention I didn't do it manually as mentioned above.

[mc-51126-boat-arrow-fix.patch](https://github.com/user-attachments/files/30679238/mc-51126-boat-arrow-fix.patch)

anzz1:
What do you mean passenger? In 1.8 vehicles can only take 1 person, no?

The upside of the tick based solution is that its *much* more performant, I'm not saying though that the perf hit of doing the collision check is necessarily meaningful though. Considering that projectiles have minimum speed, and the possible vehicles and their sizes are known beforehand, its possible to guarantee that the projectile has left the collision by ticks alone.
I'm inclined to keep it as-is, unless a bug in this approach can be demonstrated. I haven't been able to make this patch trip up in my testing, even without changing the default 5 and 25 cut offs that were meant for just the player.

Sure the Mojang's solution of doing the collision check is better in the future proof kind of way, that it will work with different future vehicles and their sizes. But we have the advantage of designing for only 1.8 here without needing to worry about future systems, giving us the chance to take performance by using shortcuts.

uRyanxD:
> What do you mean passenger? In 1.8 vehicles can only take 1 person, no?

Oh, that's right—I forgot that version 1.8 didn't support two people in the vehicles.

patches/server/0134-Projectile-Fixes.patch
```diff
@@ -0,0 +1,70 @@
...
+diff --git a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
+index fc00a5f67c5a6b2642c23834da1bef31338e2f07..64384362774ed9b7f06f4ea880249ccc34cadbc3 100644
+--- a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
++++ b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
+@@ -70,4 +70,6 @@ public class PandaSpigotWorldConfig {
+         "behaviour while still providing performance optimizations.\n" +
+         "Default: false")
+     public boolean useSpigotChunkTickingSelection = false;
++
++    public boolean projectileFixes = true;
```
uRyanxD: We don't need a config for this.

patches/server/0134-Projectile-Fixes.patch
```diff
@@ -0,0 +1,70 @@
...
+diff --git a/src/main/java/net/minecraft/server/EntityFireball.java b/src/main/java/net/minecraft/server/EntityFireball.java
+index 2524e2b8d8e03e4c97464eb74a964b595584f8a7..662a4c132a9bc1b332b7b42b7c1d1a47754dfa49 100644
+--- a/src/main/java/net/minecraft/server/EntityFireball.java
++++ b/src/main/java/net/minecraft/server/EntityFireball.java
+@@ -106,7 +106,7 @@ public abstract class EntityFireball extends Entity {
+             for (int i = 0; i < list.size(); ++i) {
+                 Entity entity1 = (Entity) list.get(i);
+ 
+-                if (entity1.ad() && (!entity1.k(this.shooter) || this.as >= 25)) {
++                if (entity1.ad() && ((!entity1.k(this.shooter) && (!this.world.pandaSpigotConfig.projectileFixes || entity1.passenger != this.shooter)) || this.as >= 25)) { // PandaSpigot - Projectile Fixes
```
uRyanxD: Why apply this to fireballs? A player can't throw them; I tested it with a Blaze in a minecart, and it seems to work fine in vanilla—it should work well with Ghasts and Withers too.

Edit: It works well with Ghasts too; the Wither refuses to get into the minecart, so I can't test it with him.

patches/server/0134-Projectile-Fixes.patch
```diff
@@ -0,0 +1,70 @@
...
+diff --git a/src/main/java/net/minecraft/server/EntityFishingHook.java b/src/main/java/net/minecraft/server/EntityFishingHook.java
+index 6eca198f9f026c82b8b31d446c3953dabdc956cf..bf3df19fdfe9643f8d78a12d169ece561e84da6c 100644
+--- a/src/main/java/net/minecraft/server/EntityFishingHook.java
++++ b/src/main/java/net/minecraft/server/EntityFishingHook.java
+@@ -168,7 +168,7 @@ public class EntityFishingHook extends Entity {
+             for (int i = 0; i < list.size(); ++i) {
+                 Entity entity1 = (Entity) list.get(i);
+ 
+-                if (entity1.ad() && (entity1 != this.owner || this.au >= 5)) {
++                if (entity1.ad() && ((entity1 != this.owner && (!this.world.pandaSpigotConfig.projectileFixes || entity1.passenger != this.owner)) || this.au >= 5)) { // PandaSpigot - Projectile Fixes
```
uRyanxD: I couldn't reproduce the bug with the fishing rod; is this diff really necessary?

patches/server/0134-Projectile-Fixes.patch
anzz1: Can you put them in a boat? Boats are the ones that i had all the issues with.
Really its about due diligence, maybe someone use a mod that put a blaze on a boat, or allow player to shoot fireball.
Essentially wherever mojang logic existed of excluding the shooter from hitting his own projectile, i added the part of excluding the ridden entity too.

patches/server/0134-Projectile-Fixes.patch
anzz1: refer https://github.com/hpfxd/PandaSpigot/pull/421#discussion_r3741442091

patches/server/0134-Projectile-Fixes.patch
anzz1: Hmmm, true, I guess we don't. I'll remove that. It can be re added later if there is some other actually controversial projectile mechanic changes.

patches/server/0134-Projectile-Fixes.patch
uRyanxD: From what I've tested, it's not possible; I believe placing entities in boats is only possible in version 1.9

uRyanxD (approved):
LGTM

PR #422, opened 2026-06-29
Cactus Fix

anzz1:
Cactus no longer damages dropped cactus blocks

### DIFF
--- a/patches/server/0137-Cactus-Fix.patch
+++ b/patches/server/0137-Cactus-Fix.patch
@@ -1,0 +1,20 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Mon, 29 Jun 2026 14:12:17 +0000
+Subject: [PATCH] Cactus Fix
+
+Cactus no longer damages dropped cactus blocks
+
+diff --git a/src/main/java/net/minecraft/server/EntityItem.java b/src/main/java/net/minecraft/server/EntityItem.java
+index 47272471c8477cb34720b116c212cd3f61f4735f..cd4dea6246290be970a70c41fa02039cf68e58a5 100644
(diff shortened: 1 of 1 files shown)


uRyanxD:
The item isn't removed anymore, but it becomes invisible :p

anzz1:
Yeah it looks like the cactus block eats it like the item goes inside the block i don't really understand why
There shouldn't be any other interaction after canceling the damage one, i think it has to do with the special bounding box it has and the type of EnumWorldBlockLayer.CUTOUT instead of regular SOLID or TRANSPARENT and that messes with the rendering somehow idk

anzz1:
This one doesn't actually need to be added to the server itself, as it can be done easily with a plugin.
```java
package com.example.cactusfix;

import org.bukkit.Material;
import org.bukkit.block.Block;
import org.bukkit.entity.Entity;
import org.bukkit.entity.Item;
import org.bukkit.event.EventHandler;
import org.bukkit.event.Listener;
import org.bukkit.event.entity.EntityDamageByBlockEvent;
import org.bukkit.event.entity.EntityDamageEvent.DamageCause;
import org.bukkit.plugin.java.JavaPlugin;

public class CactusFix extends JavaPlugin implements Listener {

    @Override
    public void onEnable() {
        getServer().getPluginManager().registerEvents(this, this);
    }

    @EventHandler
    public void onEntityDamageByBlockEvent(EntityDamageByBlockEvent event) {
        if (event.getCause() != DamageCause.CONTACT) return;
        Block damager = event.getDamager();
        if (damager == null || damager.getType() != Material.CACTUS) return;
        Entity damagee = event.getEntity();
        if (!(damagee instanceof Item)) return;
        if (((Item)damagee).getItemStack().getType() != Material.CACTUS) return;
        event.setCancelled(true);
    }
}
```
[CactusFix.jar.zip](https://github.com/user-attachments/files/30945034/CactusFix.jar.zip)

I don't know why I didn't realize it before.

PR #424, opened 2026-06-30
Configurable Soil Trampling

anzz1:
Whether players and mobs can destroy crops by jumping on them.

### DIFF
--- a/patches/server/0136-Configurable-Soil-Trampling.patch
+++ b/patches/server/0136-Configurable-Soil-Trampling.patch
@@ -1,0 +1,33 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Fri, 3 Jul 2026 01:28:19 +0000
+Subject: [PATCH] Configurable Soil Trampling
+
+Whether players and mobs can destroy crops by jumping on them.
+
+diff --git a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
+index 64384362774ed9b7f06f4ea880249ccc34cadbc3..a45c7fcd8aace57ac7408e691d88b8f249b9c81d 100644
+--- a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
++++ b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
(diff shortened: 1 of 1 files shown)


hpfxd:
This should already be covered by existing API -- not sure if it makes sense to include a config option when it can be handled by a plugin in a standard way

[example](https://github.com/KettleMC-Network/SimpleNoCropTrample/blob/3ac0710cdfce6b88aae57d0ea36da9b4db64d923/src/net/kettlemc/nocroptrample/listener/TrampleListener.java#L21-L41)

uRyanxD:
It is also possible via the EntityChangeBlockEvent

anzz1:
> This should already be covered by existing API -

Indeed it seems that using `EntityInteractEvent` and filtering for `Material.EARTH` could work as it is the only way "interact" is called here. The name threw me off I thought "Interact" would be any interaction like clicking the block or whatever but it seems not, that its only these very few specific interactions that it covers. In later Spigot it has been extended to cover much more blocks (in 1.8 it doesn't even cover Doors, really?) but still the same logic remain for dirt blocks falling on them is still the only interaction. So yeah, this is probably redundant, even if more interaction events are backported from the later versions, the dirt block interaction blocking would still work for stopping trampling and nothing else.

> It is also possible via the EntityChangeBlockEvent

This could be called from other places too like Enderman picking up a block. I guess if you filtered out entityender and entitywither then it could work, but there is also PathFinderGoalEatTile, whatever that is, that calls the changeblockevent yet no easy way to discern that call from the fall one.

anzz1:

uRyanxD:
> This could be called from other places too like Enderman picking up a block. I guess if you filtered out entityender and entitywither then it could work, but there is also PathFinderGoalEatTile, whatever that is, that calls the changeblockevent yet no easy way to discern that call from the fall one.

```java
    @EventHandler
    public void onEntityChangeBlock(EntityChangeBlockEvent event) {
        if (event.getBlock().getType() == Material.SOIL && event.getTo() == Material.DIRT) {
            event.setCancelled(true);
        }
    }
```

it works fine for me

anzz1:
Ohhh that is indeed right I somehow missed the part that EntityChangeBlockEvent contains both *from* and *to* parts, and nothing else but trampling will turn `SOIL` to `DIRT`.

This means this PR isn't needed, as even if EntityInteractEvent is updated to backport new interactions and isn't necessarily the correct one to use, the EntityChangeBlockEvent on the other hand will never have anything else turning soil to dirt so it will always be correct.

anzz1:
Here is "No Soil Trampling" as a plugin for whoever needs it.

[NoSoilTrampling.jar.zip](https://github.com/user-attachments/files/30294630/NoSoilTrampling.jar.zip)

PR #429, opened 2026-07-02
Starlight

anzz1:
Implement spottedleaf's Starlight optimized lighting engine
This vastly improves lighting performance and fixes vanilla lighting bugs  
Random light updates to "fix" broken lighting are no longer required at all

Ported from the Minecraft 1.7.3 beta version from: https://github.com/Spottedleaf/canyon/blob/master/CraftBukkit-Patches/0031-Starlight.patch

Everything seems to work fine and I couldn't find any lighting errors looking around but it needs further testing that everything is actually right, and also make sure that I caught and stopped all the vanilla light engine updates from happening when Starlight is enabled.

More information on the technical details of the engine can be found here: https://github.com/PaperMC/Starlight/blob/fabric/TECHNICAL_DETAILS.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,9 @@ Fork of Paper for 1.8.8 focused on improved performance and stability.
     - **Improved flush handling** to massively improve entity tracker performance.
     - **Support for Unix domain sockets** to avoid the overhead of TCP when using a proxy on the same machine.
 
+- **Faster and correct lighting**
+    - Implemented **[Starlight](https://github.com/PaperMC/Starlight)** lighting engine, which greatly improves lighting performance and fixes vanilla lighting bugs.
+
 - **More configuration options**, such as:
     - Customizable knockback
     - World and player data saving
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ Normal commands:
 | `setup` | Remap, decompile, and patch Minecraft. Can be run from anywhere. |
 | `p`, `patch` | Apply all patches to the project without building it. Can be run from anywhere. |
 | `j`, `jar` | Apply all patches and build the project, `paperclip.jar` will be output. Can be run from anywhere. |
-| `c`, `clean` | Removes all generated files, `PandaSpigot-API`, `PandaSpigot-Server`, and `work`. |
+| `c`, `clean` | Removes build files under `PandaSpigot-API`, `PandaSpigot-Server`. |
+| `dc`, `distclean` | Removes all generated files, `PandaSpigot-API`, `PandaSpigot-Server`, and work. |
 | `con`, `continue` | Shortcut command for running `git am --continue` or `git rebase --continue`. |
 
 Commands that require `. ./panda install` first:
--- a/panda
+++ b/panda
@@ -92,10 +92,15 @@ case "$1" in
         cd "$basedir/PandaSpigot-Server"
     ;;
     "c" | "clean")
+        rm -rf PandaSpigot-API/build
+        rm -rf PandaSpigot-Server/build
+        echo "Cleaned PandaSpigot build files"
+    ;;
+    "dc" | "distclean")
         rm -rf PandaSpigot-API
         rm -rf PandaSpigot-Server
         rm -rf base
-        echo "Cleaned build files"
+        echo "Cleaned ALL generated files"
     ;;
     "con" | "continue")
         if [ -d ".git/rebase-apply" ]; then
@@ -172,7 +177,8 @@ case "$1" in
         echo "  * setup               | Remap, decompile, and patch Minecraft. Can be run from anywhere."
         echo "  * p, patch            | Apply all patches to the project without building it. Can be run from anywhere."
         echo "  * j, jar              | Apply all patches and build the project, paperclip.jar will be output. Can be run from anywhere."
-        echo "  * c, clean            | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, and work."
+        echo "  * c, clean            | Removes build files under PandaSpigot-API, PandaSpigot-Server."
+        echo "  * dc, distclean       | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, and work."
         echo "  * con, continue       | Shortcut command for running git am --continue, or git rebase --continue."
         echo ""
         echo " These commands require the install command before use:"
--- a/patches/server/0137-Starlight.patch
+++ b/patches/server/0137-Starlight.patch
@@ -1,0 +1,941 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Fri, 3 Jul 2026 02:14:47 +0000
+Subject: [PATCH] Starlight
+
+Implement Starlight optimized lighting engine
+This vastly improves lighting performance and fixes vanilla lighting bugs
+Random light updates to "fix" broken lighting are no longer required at all
+
+diff --git a/src/main/java/ca/spottedleaf/starlight/StarlightEngine.java b/src/main/java/ca/spottedleaf/starlight/StarlightEngine.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..8b163545a5cd0de69015af390eaff4c57839c37d
+--- /dev/null
++++ b/src/main/java/ca/spottedleaf/starlight/StarlightEngine.java
+@@ -0,0 +1,710 @@
++package ca.spottedleaf.starlight;
++
++import net.minecraft.server.Block;
++import net.minecraft.server.BlockPosition;
++import net.minecraft.server.Blocks;
++import net.minecraft.server.Chunk;
++import net.minecraft.server.EnumSkyBlock;
++import net.minecraft.server.IBlockData;
++import net.minecraft.server.World;
++
++import java.util.ArrayList;
++import java.util.Arrays;
++import java.util.List;
++
++public class StarlightEngine {
++
++    protected static enum AxisDirection {
++
++        // Declaration order is important and relied upon. Do not change without modifying propagation code.
++        POSITIVE_X(1, 0, 0), NEGATIVE_X(-1, 0, 0),
++        POSITIVE_Z(0, 0, 1), NEGATIVE_Z(0, 0, -1),
++        POSITIVE_Y(0, 1, 0), NEGATIVE_Y(0, -1, 0);
++
++        static {
++            POSITIVE_X.opposite = NEGATIVE_X; NEGATIVE_X.opposite = POSITIVE_X;
++            POSITIVE_Z.opposite = NEGATIVE_Z; NEGATIVE_Z.opposite = POSITIVE_Z;
++            POSITIVE_Y.opposite = NEGATIVE_Y; NEGATIVE_Y.opposite = POSITIVE_Y;
++        }
++
++        protected AxisDirection opposite;
++
++        public final int x;
++        public final int y;
++        public final int z;
++        public final long everythingButThisDirection;
++        public final long everythingButTheOppositeDirection;
++
++        AxisDirection(final int x, final int y, final int z) {
++            this.x = x;
++            this.y = y;
++            this.z = z;
++            this.everythingButThisDirection = (long)(ALL_DIRECTIONS_BITSET ^ (1 << this.ordinal()));
++            // positive is always even, negative is always odd. Flip the 1 bit to get the negative direction.
++            this.everythingButTheOppositeDirection = (long)(ALL_DIRECTIONS_BITSET ^ (1 << (this.ordinal() ^ 1)));
++        }
++
++        public AxisDirection getOpposite() {
++            return this.opposite;
++        }
++    }
++
++    protected static final AxisDirection[] DIRECTIONS = AxisDirection.values();
++    protected static final AxisDirection[] AXIS_DIRECTIONS = DIRECTIONS;
++    protected static final AxisDirection[] ONLY_HORIZONTAL_DIRECTIONS = new AxisDirection[] {
++            AxisDirection.POSITIVE_X, AxisDirection.NEGATIVE_X,
++            AxisDirection.POSITIVE_Z, AxisDirection.NEGATIVE_Z
++    };
++
++    protected static final AxisDirection[][] OLD_CHECK_DIRECTIONS = new AxisDirection[1 << 6][];
++    protected static final int ALL_DIRECTIONS_BITSET = (1 << 6) - 1;
++    static {
++        for (int i = 0; i < OLD_CHECK_DIRECTIONS.length; ++i) {
++            final List<AxisDirection> directions = new ArrayList<>();
++            for (int bitset = i, len = Integer.bitCount(i), index = 0; index < len; ++index, bitset ^= (-bitset & bitset)) {
++                directions.add(AXIS_DIRECTIONS[Integer.numberOfTrailingZeros(bitset)]);
++            }
++            OLD_CHECK_DIRECTIONS[i] = directions.toArray(new AxisDirection[0]);
++        }
++    }
++
++    // always initialised during start of lighting.
++    // index = x + (z * 5)
++    protected final Chunk[] chunkCache = new Chunk[5 * 5];
++
++    protected int encodeOffsetX;
++    protected int encodeOffsetY;
++    protected int encodeOffsetZ;
++
++    protected int coordinateOffset;
++
++    protected int chunkOffsetX;
++    protected int chunkOffsetY;
++    protected int chunkOffsetZ;
++
++    protected int chunkIndexOffset;
++    protected int chunkSectionIndexOffset;
++
++    protected final boolean skylightPropagator;
++    protected final int emittedLightMask;
++    protected final boolean isClientSide;
++
++    protected final World world;
++
++    public StarlightEngine(final boolean skylightPropagator, final World world) {
++        this.skylightPropagator = skylightPropagator;
++        this.emittedLightMask = skylightPropagator ? 0 : 0xF;
++        this.isClientSide = false;
++        this.world = world;
++    }
++
++    protected final void setupEncodeOffset(final int centerX, final int centerY, final int centerZ) {
++        // 31 = center + encodeOffset
++        this.encodeOffsetX = 31 - centerX;
++        this.encodeOffsetY = (-(-1 - 1) << 4); // we want 0 to be the smallest encoded value
++        this.encodeOffsetZ = 31 - centerZ;
++
++        // coordinateIndex = x | (z << 6) | (y << 12)
++        this.coordinateOffset = this.encodeOffsetX + (this.encodeOffsetZ << 6) + (this.encodeOffsetY << 12);
++
++        // 2 = (centerX >> 4) + chunkOffset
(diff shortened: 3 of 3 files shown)


anzz1:
This is the known good StarlightEngine.java from the Beta 1.7.3 version:
[StarLightEngine.java](https://github.com/user-attachments/files/29609489/StarLightEngine.java)
for posterity and diffing purposes for comparing to the ported one.

patches/server/0138-Starlight.patch
```diff
@@ -0,0 +1,949 @@
...
+             chunk = this.loadChunk(i, j);
++            boolean generated = chunk == null; // PandaSpigot - Starlight
+             if (chunk == null) {
+                 if (this.chunkProvider == null) {
+                     chunk = this.emptyChunk;
+@@ -178,6 +179,7 @@ public class ChunkProviderServer implements IChunkProvider {
+             this.chunks.put(key, chunk); // PandaSpigot - Reuse key
+             
+             chunk.addEntities();
++            if (chunk.world.pandaSpigotConfig.useStarlight && generated) chunk.initLightingForRealNotJustHeightmap();
```
uRyanxD: Can't we use the existing `newChunk` variable?

anzz1:

patches/server/0138-Starlight.patch
anzz1: We can indeed, I'll fix that at the same time as I rebase the config later. edit: done

anzz1:
Actually I'm not sure whether this
```diff
-if (PlayerChunkMap.this.world.spigotConfig.randomLightUpdates && !chunk.u() && chunk.isDone()) { // PandaSpigot - Starlight
+if ((PlayerChunkMap.this.world.pandaSpigotConfig.useStarlight || PlayerChunkMap.this.world.spigotConfig.randomLightUpdates) && !chunk.u() && chunk.isDone()) { // PandaSpigot - Starlight`
```
in the ACTS is correct, and actually probably shouldn't be there as chunk ticks have nothing to do with the lighting anymore.

In Starlight, whether chunk is "lit" or not is irrelevant as such property is no longer acknowledged, the chunks are always lighted once, correctly, when the chunks are generated and never again, and lighting is only updated when block state is changed for that particular block and whatever propagates from it, as it should be. There is no such thing as "unlit" state anymore. At least this is my understanding of the system.

anzz1:
And considering this whole system is about doing the lighting correctly once when the chunks are generated, if you use this on a pre-existing world it doesn't fix existing lighting errors as the whole point is to not do such things, as the lighting is never supposed to go into an incorrect state. The system is now stateless, it inits the lighting once, and then updates only the lights needed on block state change, there is no lighting states or ticks or updates anymore, just the saved light values. In fact half of the functions in chunk.java become dead code with Starlight as none of the lighting related functions are necessary anymore.


However many people will probably want to switch to using Starlight on their existing worlds, so maybe implementing a console command that relights all saved chunks with Starlight which you'd need to do once when migrating to it would be advisable.

anzz1:
Alright I re-used the `newChunk` as discussed, and negated the logic in ACTS that lighting is *not* taken into account when deciding which chunks to tick
`if (PlayerChunkMap.this.world.spigotConfig.randomLightUpdates && !PlayerChunkMap.this.world.pandaSpigotConfig.useStarlight && !chunk.u() && chunk.isDone()) { // PandaSpigot - Starlight`
that's how it was supposed to be in the first place I had some brain fart there.

anzz1:
I made this relight command which can be used to relight existing worlds with Starlight. 
It is MCP/vanilla version that just needs to be ported, it does ~200chunks/sec on vanilla (if pandaspigot has optimized the chunk loader it could be faster), its slower than i'd like but then again it only needs to be run once after enabling starlight and never again. The command relights everything, all dimensions and all chunks, in memory or on disk, loading and unloading chunks as it goes.

Not sure if it should be incorporated with the starlight engine or should it be a plugin instead?

It could probably be optimized further like reveal the `id2ChunkMap` from ChunkProviderServer and iterate on those packed XZ longs instead of Chunks in the `unloadOutsideRadius` function, and implement some better algorithm of figuring which chunks to unload when the spiral is walked than the simple radius check which results in unloading and loading same chunks. I also considered BFS walking which would reduce duplicate unload-loads but I wasn't certain whether the result would be correct like it is for the outward spiral walk, and again as it needs to be done once simplicity and correctness was the main goal here, not maximum performance.

edit: i tried BFS+spiral (use BFS first, then spiral order to create new BFS starting points for chunks that weren't visited by BFS starting from spawn) but the BFS+spiral made it slower than just spiral, at ~160 chunk/sec vs just spiral's ~200.

```java
package net.minecraft.command;

import net.minecraft.server.MinecraftServer;
import net.minecraft.util.BlockPos;
import net.minecraft.util.IProgressUpdate;
import net.minecraft.world.ChunkCoordIntPair;
import net.minecraft.world.MinecraftException;
import net.minecraft.world.World;
import net.minecraft.world.WorldServer;
import net.minecraft.world.chunk.Chunk;
import net.minecraft.world.chunk.storage.RegionFile;
import net.minecraft.world.chunk.storage.RegionFileCache;
import net.minecraft.world.gen.ChunkProviderServer;

import java.io.File;
import java.io.IOException;
import java.util.ArrayList;
import java.util.HashSet;
import java.util.List;
import java.util.Set;

public class CommandRelight extends CommandBase
{
    public String getCommandName()
    {
        return "relight";
    }

    public int getRequiredPermissionLevel()
    {
        return 2;
    }

    public String getCommandUsage(ICommandSender sender)
    {
        return "/relight all";
    }

    public void processCommand(ICommandSender sender, String[] args) throws CommandException
    {
        if (args.length != 1 || !args[0].equalsIgnoreCase("all"))
        {
            throw new WrongUsageException("/relight all", new Object[0]);
        }

        try {
            MinecraftServer minecraftserver = MinecraftServer.getServer();

            notifyOperators(sender, this, "Relight ALL start", new Object[0]);
            long ttime = System.currentTimeMillis();

            for (int j = 0; j < minecraftserver.worldServers.length; ++j) {
                if (minecraftserver.worldServers[j] == null) continue;

                WorldServer worldserver = minecraftserver.worldServers[j];
                ChunkProviderServer chunkProvider = (ChunkProviderServer)worldserver.getChunkProvider();

                Set<Long> allChunks = new HashSet<>();

                File regionDir = new File(getDimensionDirectory(worldserver.getSaveHandler().getWorldDirectory(), worldserver.provider.getDimensionId()), "region");
                File[] regionFiles = regionDir.listFiles((dir, name) -> name.endsWith(".mca"));

                for (File file : regionFiles) {
                    String[] parts = file.getName().split("\\.");
                    int regionX = Integer.parseInt(parts[1]);
                    int regionZ = Integer.parseInt(parts[2]);

                    RegionFile region = RegionFileCache.createOrLoadRegionFile(getDimensionDirectory(worldserver.getSaveHandler().getWorldDirectory(), worldserver.provider.getDimensionId()), (regionX << 5), (regionZ << 5));
                    for (int cx = 0; cx < 32; cx++) {
                        for (int cz = 0; cz < 32; cz++) {
                            if (region.isChunkSaved(cx, cz)) {
                                int chunkX = (regionX << 5) + cx;
                                int chunkZ = (regionZ << 5) + cz;
                                allChunks.add(ChunkCoordIntPair.chunkXZ2Int(chunkX, chunkZ));
                            }
                        }
                    }
                }

                Set<Long> originallyLoaded = new HashSet<>();
                for (Chunk chunk : chunkProvider.getLoadedChunks()) {
                    originallyLoaded.add(ChunkCoordIntPair.chunkXZ2Int(chunk.xPosition, chunk.zPosition));
                }
                allChunks.addAll(originallyLoaded);

                if (allChunks.size() == 0) continue;

                notifyOperators(sender, this, "Relighting dimension " + worldserver.provider.getDimensionId() + " \'" + worldserver.provider.getDimensionName() + "\'", new Object[0]);
                notifyOperators(sender, this, "Relighting " + allChunks.size() + " chunks (" + originallyLoaded.size() + " loaded, " +
                    (allChunks.size() - originallyLoaded.size()) + " unloaded)", new Object[0]);
                long dtime = System.currentTimeMillis();

                BlockPos spawn = worldserver.getSpawnPoint();
                final int sx = spawn.getX() >> 4;
                final int sz = spawn.getZ() >> 4;

                List<Long> ordered = new ArrayList<>(allChunks);
                ordered.sort((a, b) -> {
                    int ax = (int)(a & 4294967295L) - sx;
                    int az = (int)((a >>> 32) & 4294967295L) - sz;
                    int bx = (int)(b & 4294967295L) - sx;
                    int bz = (int)((b >>> 32) & 4294967295L) - sz;

                    // 1. distance (square radius)
                    int da = Math.max(Math.abs(ax), Math.abs(az));
                    int db = Math.max(Math.abs(bx), Math.abs(bz));

                    if (da != db) return Integer.compare(da, db);

                    // 2. spiral ordering within the ring
                    // map position to "edge order" around square
                    int oa = spiralIndex(ax, az, da);
                    int ob = spiralIndex(bx, bz, db);

                    return Integer.compare(oa, ob);
                });

                for (long key : ordered) {
                    int x = (int)(key & 4294967295L);
                    int z = (int)((key >>> 32) & 4294967295L);

                    boolean wasLoaded = chunkProvider.chunkExists(x, z);
                    Chunk chunk = wasLoaded
                        ? chunkProvider.provideChunk(x, z)
                        : chunkProvider.loadChunkNoGenerate(x, z);

                    if (chunk == null) continue;

                    // ensure neighbors (radius 1)
                    for (int dx = -1; dx <= 1; dx++) {
                        for (int dz = -1; dz <= 1; dz++) {
                            int nx = x + dx;
                            int nz = z + dz;

                            long nKey = ChunkCoordIntPair.chunkXZ2Int(nx, nz);
                            if (!allChunks.contains(nKey)) continue;

                            if (!chunkProvider.chunkExists(nx, nz)) {
                                chunkProvider.loadChunkNoGenerate(nx, nz);
                            }
                        }
                    }

                    // relight
                    worldserver.blockLight.initBlockLight(x, z);
                    if (!worldserver.provider.getHasNoSky()) {
                        worldserver.skyLight.initSkylight(x, z);
                    }

                    chunk.setModified(true);

                    // unload outside radius 2
                    unloadOutsideRadius(chunkProvider, originallyLoaded, x, z, 2);
                }

                for (long key : allChunks) {
                    if (chunkProvider.chunkExists(key) && !originallyLoaded.contains(key)) {
                        chunkProvider.dropChunk((int)(key & 4294967295L), (int)((key >>> 32) & 4294967295L));
                    }
                }
                chunkProvider.unloadQueuedChunks();

                notifyOperators(sender, this, "Relighting dimension " + worldserver.provider.getDimensionId() + " done, took " + ((System.currentTimeMillis() - dtime) / 1000) + " seconds (" +
                    ((allChunks.size() * 1000) / (System.currentTimeMillis() - dtime)) + " chunks/s)", new Object[0]);

                if (chunkProvider.getLoadedChunkCount() > 0) {
                    notifyOperators(sender, this, "Saving all " + chunkProvider.getLoadedChunkCount() + " loaded chunks", new Object[0]);

                    try
                    {
                        worldserver.saveAllChunks(true, (IProgressUpdate)null);
                    }
                    catch (MinecraftException minecraftexception)
                    {
                        minecraftexception.printStackTrace();
                        notifyOperators(sender, this, "Saving chunks failed!", new Object[0]);
                        return;
                    }
               }
            }
            notifyOperators(sender, this, "Relight ALL done! Took " + ((System.currentTimeMillis() - ttime) / 1000) + " seconds total", new Object[0]);
        }
        catch (Exception ex)
        {
            ex.printStackTrace();
            throw ex;
        }
    }

    private File getDimensionDirectory(File worldDir, int dimensionId) {
        if (dimensionId == 0) {
          return worldDir;
        } else {
          return new File(worldDir, "DIM" + dimensionId);
        }
    }

    private void unloadOutsideRadius(ChunkProviderServer chunkProvider, Set<Long> originallyLoaded, int centerX, int centerZ, int radius) {
        for (Chunk chunk : chunkProvider.getLoadedChunks()) {
            int dx = Math.abs(chunk.xPosition - centerX);
            int dz = Math.abs(chunk.zPosition - centerZ);

            if (dx > radius || dz > radius) {
                long key = ChunkCoordIntPair.chunkXZ2Int(chunk.xPosition, chunk.zPosition);
                if (!originallyLoaded.contains(key)) {
                    chunkProvider.dropChunk(chunk.xPosition, chunk.zPosition);
                }
            }
        }
        chunkProvider.unloadQueuedChunks();
    }

    private int spiralIndex(int x, int z, int r) {
        if (r == 0) return 0;
        if (x == r && z > -r) {
            return z + r - 1;
        } else if (z == r) {
            return (2 * r - 1) + (r - x);
        } else if (x == -r) {
            return (4 * r - 1) + (r - z);
        } else { // z == -r
            return (6 * r - 1) + (x + r);
        }
    }
}
```

killiannnnnn:
any updates ?

anzz1:
Everything is working fine as far as I can tell; you can pull [my master branch](https://github.com/anzz1/PandaSpigot/tree/master) that has the PRs waiting for review merged and test out for yourself if you'd like.

anzz1:
Testing this on vanilla, I found out that I needed to add
`this.world.markBlockForUpdate(new BlockPos(worldX, worldY, worldZ));` aka `this.world.notify(new BlockPosition(worldX, worldY, worldZ));`
to the `postLightUpdate` as otherwise sometimes the client wouldn't receive a light update that had propagated to a neighbouring chunk.
However I haven't been able to replicate this bug on PandaSpigot so possibly some another patch fixed this issue and simply `setLightFor` aka `setBrightness` is enough to mark the block/chunk dirty and have the update sent to the client.

edit: actually, nevermind, the bug did exist here too, I just needed to recreate the exact conditions i tested in vanilla to coax it out. now it works properly after marking the blocks dirty in `postLightUpdate`

uRyanxD:
Shouldn't `World#setTypeAndData` replace the `x(BlockPos)` call—which would be something like `checkLight(BlockPos)`—with a Starlight-specific version?

Something like:

```java
   public boolean checkLight(BlockPos pos) {
      this.blockLight.checkBlockEmittance(pos.getX(), pos.getY(), pos.getZ());
      if (!this.dimension.hasNoSky()) {
         this.skyLight.checkSkyEmittance(pos.getX(), pos.getY(), pos.getZ());
      }
      return true;
   }
```

anzz1:
@uRyanxD:
It happens later down the call stack: World.setTypeAndData -> Chunk.a -> Chunk.updateLight
But yeah this PR should definitely be scrutinized closely - I'm not 100% certain of what I'm doing here I don't know the intricacies of the lighting system I just ported Starlight for the Beta version without really changing anything and I might have missed something.
However after adding the postLightUpdate fix in the latest commit f8887ee , I haven't been able to find any errors in my testing.

anzz1:
And the `x` call ends up ultimately to -> `updateLight` -> `updateBrightness` -> `c` where just `true` is returned if starlight is used.

uRyanxD:
> And the `x` call ends up ultimately to -> `updateLight` -> `updateBrightness` -> `c` where just `true` is returned if starlight is used.

This could leave the block with an old lighting value, if we used /setblock or the Bukkit API, for example? (I haven't tested it and haven't looked deeply into the code, it's just a guess)

anzz1:
> This could leave the block with an old lighting value, if we used /setblock or the Bukkit API, for example?

Oh yeah I haven't followed these cases nor the possible cases how bukkit could call something maybe bypassing starlight, i've only tested regular vanilla gameplay so far essentially just playing and flying around and testing other patches and stuff and I haven't seen any lighting errors.
But as long as the `World.setTypeandData` aka `World.setBlockState` is being called it should result to updating the light too and everything working fine. /setblock at least in vanilla does unless bukkit changed it not to do that.

uRyanxD:
@anzz1 You accidentally merged the changes from PR https://github.com/hpfxd/PandaSpigot/pull/433 into this one

anzz1:
Ah yeah, it seems so, the readme has also changed since this was submitted. The chronological order isn't quite right anymore either e.g. the latest PR I submitted #445 is on top of the current master and not in the "PR queue" order.
I think the easiest way for now is to go through the queue skipping the PRs that don't merge cleanly and I'll then go through the skipped ones rebasing to the master and put them on top of that in correct chronological order.

overkidding:
btw this should use paper async lightning when enabled, if not, in a normal generated world (not void or flat) i get up to 200mspt and tps drop significantly

anzz1:
> btw this should use paper async lightning when enabled

Afaik the async lighting code is buggy and paper removed it completely on later version. Unless there is some que for light updates and some changes are made so chunk saves won't happen when something has async light update que'd, requiring the chunk saves to be que'd too, and what if situation changes while the que hasn't reached it yet, and how about change happening in middle of propagation, etc. there is a lot of problems to solve if went the async route. This is probably why they removed the async code altogether because to make it properly you cant half ass it and it require lot of changes. Its def easier and probably even faster to just optimize the light updates than trying to make a properly functioning async engine for block changes.

> in a normal generated world (not void or flat) i get up to 200mspt and tps drop significantly

Strange, I haven't have any perf problems and in fact the perf of this should be much better than vanilla light engine, according to the technical document at least. Although considering I converted this from spottedleaf's code myself without having a full understanding of the entire lighting engine, it is certainly possible that i'm doing some extra calculations i shouldn't, like maybe there are still some parts of the vanilla lighting engine running that shouldnt be, creating extra work.

Also I only ever tested playing by myself, so I dont really see performance problem with any light engine, so I assume you have lots of players and then this problem manifests itself?

overkidding:
> Strange, I haven't have any perf problems and in fact the perf of this should be much better than vanilla light engine,

I tried this by speed flying around a world with 10x fly boost, also i dont seem to have issues with async lightning, and about the newer versions is cause everything was made asynchronous starting from the world gen, so probably the lightning is managed by the same threads as the gen/load ones but certainly the lightning is not running on the main thread.

anzz1:
Ah good now I understand your simulated situation, that doesn't mean there was anything wrong with the engine, you were just creating a lot of new chunks. That is the situation where this lighting system can be slower since even if its optimized, the whole point is that there is no "lit/unlit" states anymore the lighting is always generated correct. Which mean that when you generate a new chunk, the entire lighting for it is generated as the world is. So if you're creating bunch of them at once, it indeed can create lag. The other side of it is that once its generated, it only ever updates lighting when there is a block change the small part of the chunk, and there will never be unnecessary work with "lighting fix" passes like the vanilla engine. So it will be much faster once the chunk is generated.

Not having to generate entire lighting in one go is probably why they went with the lit/unlit system in first place, in attempt to make chunk generation less intensive. That is the vanilla system it creates chunk as unlit, then uses some logic to "fix" the lighting afterwards, deferring the updates. However because it doesn't work properly it results in lighting bugs and due to the design also lots of unnecessary work.

And yes that would make sense that if worldgen is later made async then the lighting could happen async in the world gen thread too. If someone wants to port asynchronous world gen from the later versions that would be nice. But that is also what I meant by you can't half-ass it, it needs to be made properly, the whole system would need a more base overhaul to support it. Like I tried to explain, I didn't say that paper async lighting never works, just that the system can have race conditions to create problems by the nature of it, and that could only be fixed by overhaul and rethink of the whole system.

PR #432, opened 2026-07-06
Fix MC-140545 Pathfinding bias

anzz1:
This way when there are equidistant paths it will choose the one pointing towards the goal instead of the order what NodeProcessor puts the directions in the array which is N,W,E,S.

Configurable with fixPathFindingBias as some might prefer the vanilla bugged/biased pathing for whatever reason

### DIFF
--- a/patches/server/0138-Fix-MC-140545-Pathfinding-bias.patch
+++ b/patches/server/0138-Fix-MC-140545-Pathfinding-bias.patch
@@ -1,0 +1,46 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Mon, 6 Jul 2026 21:53:22 +0000
+Subject: [PATCH] Fix MC-140545 Pathfinding bias
+
+Configurable with fixPathFindingBias as some might prefer the vanilla
+bugged/biased pathing for whatever reason
+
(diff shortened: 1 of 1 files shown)


patches/server/0138-Fix-MC-140545-Pathfinding-bias.patch
```diff
@@ -0,0 +1,46 @@
...
+Configurable with fixPathFindingBias as some might prefer the vanilla
+bugged/biased pathing for whatever reason
+
+diff --git a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
+index be7fe9fd6028c1feda83124866feaccbb1f0c05e..14cc48df65a990d1cf2f49fa1bb0750f0b742053 100644
+--- a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
++++ b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
+@@ -80,4 +80,8 @@ public class PandaSpigotWorldConfig {
+     @Comment("Use Starlight optimized lighting engine.\n" +
+             "Default: true")
```
uRyanxD, line 15: The diff needs to be fixed.

You created it based on a branch where Starlight was already merged :p

anzz1: Oh yeah all the patches are applied in the chronological PR order so they don't have to be rebased multiple times.
For reference: https://github.com/anzz1/PandaSpigot
So one-by-one in order. It's no problem if some of the newer PR wait while older are are still reviewed, like Starlight is a larger review obviously as its quite big. No rush.

PR #433, opened 2026-07-06
panda script improvement

anzz1:
- old `clean` is now `distclean`, as canonically "clean" should only remove build files and not your current working state. Not fun when you don't RTFM and assume "clean" does what it usually does but nukes your work instead.

- `clean` is now a new command that just removes the build files under PandaSpigot-API/build and PandaSpigot-Server/build, leaving your working state intact, as you'd expect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,8 @@ Normal commands:
 | `setup` | Remap, decompile, and patch Minecraft. Can be run from anywhere. |
 | `p`, `patch` | Apply all patches to the project without building it. Can be run from anywhere. |
 | `j`, `jar` | Build the project, `paperclip.jar` will be output. Can be run from anywhere. |
-| `c`, `clean` | Removes all generated files, `PandaSpigot-API`, `PandaSpigot-Server`, and `work`. |
+| `c`, `clean` | Removes build files under `PandaSpigot-API`, `PandaSpigot-Server`. |
+| `dc`, `distclean` | Removes all generated files, `PandaSpigot-API`, `PandaSpigot-Server`, and work. |
 | `con`, `continue` | Shortcut command for running `git am --continue` or `git rebase --continue`. |
 
 Commands that require `. ./panda install` first:
--- a/panda
+++ b/panda
@@ -83,10 +83,15 @@ case "$1" in
         cd "$basedir/PandaSpigot-Server"
     ;;
     "c" | "clean")
+        rm -rf PandaSpigot-API/build
+        rm -rf PandaSpigot-Server/build
+        echo "Cleaned PandaSpigot build files"
+    ;;
+    "dc" | "distclean")
         rm -rf PandaSpigot-API
         rm -rf PandaSpigot-Server
         rm -rf base
-        echo "Cleaned build files"
+        echo "Cleaned ALL generated files"
     ;;
     "con" | "continue")
         if [ -d ".git/rebase-apply" ]; then
@@ -163,7 +168,8 @@ case "$1" in
         echo "  * setup               | Remap, decompile, and patch Minecraft. Can be run from anywhere."
         echo "  * p, patch            | Apply all patches to the project without building it. Can be run from anywhere."
         echo "  * j, jar              | Apply all patches and build the project, paperclip.jar will be output. Can be run from anywhere."
-        echo "  * c, clean            | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, and work."
+        echo "  * c, clean            | Removes build files under PandaSpigot-API, PandaSpigot-Server."
+        echo "  * dc, distclean       | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, and work."
         echo "  * con, continue       | Shortcut command for running git am --continue, or git rebase --continue."
         echo ""
         echo " These commands require the install command before use:"
--- a/panda
+++ b/panda
@@ -87,11 +87,12 @@ case "$1" in
         rm -rf PandaSpigot-Server/build
         echo "Cleaned PandaSpigot build files"
     ;;
-    "dc" | "distclean")
+    "dc" | "distclean" | "reset")
         rm -rf PandaSpigot-API
         rm -rf PandaSpigot-Server
         rm -rf base
-        echo "Cleaned ALL generated files"
+        rm -rf work
+        echo "Cleaned ALL generated, build and work files"
     ;;
     "con" | "continue")
         if [ -d ".git/rebase-apply" ]; then
@@ -169,7 +170,7 @@ case "$1" in
         echo "  * p, patch            | Apply all patches to the project without building it. Can be run from anywhere."
         echo "  * j, jar              | Apply all patches and build the project, paperclip.jar will be output. Can be run from anywhere."
         echo "  * c, clean            | Removes build files under PandaSpigot-API, PandaSpigot-Server."
-        echo "  * dc, distclean       | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, and work."
+        echo "  * dc, distclean       | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, base, and work."
         echo "  * con, continue       | Shortcut command for running git am --continue, or git rebase --continue."
         echo ""
         echo " These commands require the install command before use:"


uRyanxD:
Doesn't Gradle already have a tool for that?
```
./gradlew clean
./gradlew clean build
```
Or something like that?

anzz1:
I don't really know about gradle too much to have a solid answer but I *think* it also does other things like purge some package caches or what not. Just removing the folders manually can be sure that what you want – deleting the build files – is actually what happens. If the `clean` needs some gradle command added if it holds some hidden caches of the classes somewhere or something like that then by all means it can be added. But so far I've found that deleting the `build` folder where the classes lie, and killing the stupid daemon, are enough to make it rebuild correctly, whereas simply doing `gradlew build` its daemon which supposedly watches the filesystem for changes doesn't do it job and actually rebuild the correct changed files. Even if you do gradlew --stop to kill that stupid daemon and then rebuild, it still sometimes doesn't detect that a file has changed, does it not know to check "file modified" dates on the filesystem, which even C toolchains from the 1980s knew how to do?! Shit, I'm ranting again.

The important part of this PR is not the addition of the new `clean` command but moving the current `clean` to `distclean` to signify that its a nuclear option that deletes your work which isn't what `clean` is canonically supposed to do. You don't want to run `clean` fully expecting that you can run `build` after, only to find out it having deleted all your work.

uRyanxD:
> I don't really know about gradle too much to have a solid answer but I _think_ it also does other things like purge some package caches or what not. Just removing the folders manually can be sure that what you want – deleting the build files – is actually what happens. If the `clean` needs some gradle command added if it holds some hidden caches of the classes somewhere or something like that then by all means it can be added. But so far I've found that deleting the `build` folder where the classes lie, and killing the stupid daemon, are enough to make it rebuild correctly, whereas simply doing `gradlew build` its daemon which supposedly watches the filesystem for changes doesn't do it job and actually rebuild the correct changed files. Even if you do gradlew --stop to kill that stupid daemon and then rebuild, it still sometimes doesn't detect that a file has changed, does it not know to check "file modified" dates on the filesystem, which even C toolchains from the 1980s knew how to do?! Shit, I'm ranting again.
> 
> The important part of this PR is not the addition of the new `clean` command but moving the current `clean` to `distclean` to signify that its a nuclear option that deletes your work which isn't what `clean` is canonically supposed to do. You don't want to run `clean` fully expecting that you can run `build` after, only to find out it having deleted all your work.

I don't really understand your point.

What did you mean by "forces a correct recompilation"?

Also, regarding recompiling only changed files, there is a Gradle flag called "configuration cache" (or something similar) that does this, though I've never tested it with Panda. There is also a misunderstanding about the Daemon; it simply keeps the process running in the background so Gradle doesn't have to start up again every time you compile the project.

I also believe "clean" works as expected, since its purpose is to remove generated files, as stated in the description.

Furthermore, I don't understand why you'd want to delete the `build` folder; Gradle replaces the files with new ones as soon as you run the build command again, and as I mentioned above, `.\gradlew clean` already does what you want.

What we could do here is rename it from "clean" to "reset"—or the name you suggested—and correct the command description, since it currently says it resets the "work" folder, whereas in Panda it's called "base".

anzz1:
> I also believe "clean" works as expected, since its purpose is to remove generated files, as stated in the description.

No, not really. `clean` is canonically for build files, it shouldn't ever delete your work, configuration files, or such. `distclean` is for resetting the state to a clean slate position, deleting everything, including configuration etc. Of course people can make such commands do whatever they want, but that is the norm.

> and as I mentioned above, .\gradlew clean already does what you want.

Sure, but the `panda` tool is used to build, with `panda jar`. You wouldn't expect `clean` to delete your work. Providing a way to do `clean` via the panda script is good form.

> What we could do here is rename it from "clean" to "reset"—or the name you suggested

Why not both. I added "reset" as alias to "distclean".

> and correct the command description, since it currently says it resets the "work" folder, whereas in Panda it's called "base".

Right, the paper etc. stuff get downloaded to "base". There seems to be a possibility for a "work" folder too though, with the "edit" command. distclean now deletes both `base` and `work` , that should be everything.

anzz1:
> What did you mean by "forces a correct recompilation"?

That when the object files are gone, gradle is forced to do a clean recompilation. 

> there is a Gradle flag called "configuration cache" (or something similar)

Yes exactly there is a million different build systems and more often than not they always seem to have problems in recognizing what files have actually changed, I had problems before with gradle not recognizing the changes but it's by no means unique. What however always works, for any build system, is nuking the build files. No need to dig into configuration and wonder if some future update decides some configuration flag becomes "deprecated" or whatever and no longer works, or some bug gets introduced, or some random edge case makes it not work anyway. Nuking the build files however, guaranteed 100% success rate. 

> There is also a misunderstanding about the Daemon; it simply keeps the process running in the background so Gradle doesn't have to start up again every time you compile the project.

Actually I had success making gradle detect the changed files better if I killed the daemon between compilations. But then one time I had it not detect the change even after killing the daemon, so I just started doing which I should have from the beginning, just always do a clean recompile, even if it takes a bit of time. Beats tearing your hair out over a bug that you can't figure out because it wasn't introduced by you but rather a half-done compilation.

panda
```diff
@@ -83,10 +83,16 @@ case "$1" in
...
+        rm -rf PandaSpigot-API/build
+        rm -rf PandaSpigot-Server/build
+        echo "Cleaned PandaSpigot build files"
+    ;;
+    "dc" | "distclean" | "reset")
         rm -rf PandaSpigot-API
         rm -rf PandaSpigot-Server
         rm -rf base
-        echo "Cleaned build files"
+        rm -rf work
```
uRyanxD, line 94: We can simply change the `edit` command to use `base`; "work" was the name Paper used instead of "base".

panda
```diff
@@ -163,7 +169,8 @@ case "$1" in
         echo "  * setup               | Remap, decompile, and patch Minecraft. Can be run from anywhere."
         echo "  * p, patch            | Apply all patches to the project without building it. Can be run from anywhere."
         echo "  * j, jar              | Apply all patches and build the project, paperclip.jar will be output. Can be run from anywhere."
-        echo "  * c, clean            | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, and work."
+        echo "  * c, clean            | Removes build files under PandaSpigot-API, PandaSpigot-Server."
+        echo "  * dc, distclean       | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, base, and work."
```
uRyanxD, line 173:
```diff
- echo "  * dc, distclean       | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, base, and work."
+ echo "  * dc, distclean, reset       | Removes all generated files, PandaSpigot-API, PandaSpigot-Server, and base."
```

README.md
```diff
@@ -86,7 +86,8 @@ Normal commands:
 | `setup` | Remap, decompile, and patch Minecraft. Can be run from anywhere. |
 | `p`, `patch` | Apply all patches to the project without building it. Can be run from anywhere. |
 | `j`, `jar` | Build the project, `paperclip.jar` will be output. Can be run from anywhere. |
-| `c`, `clean` | Removes all generated files, `PandaSpigot-API`, `PandaSpigot-Server`, and `work`. |
+| `c`, `clean` | Removes build files under `PandaSpigot-API`, `PandaSpigot-Server`. |
+| `dc`, `distclean` | Removes all generated files, `PandaSpigot-API`, `PandaSpigot-Server`, and work. |
```
uRyanxD, line 90:
```diff
-   `dc`, `distclean` | Removes all generated files, `PandaSpigot-API`, `PandaSpigot-Server`, and work. |
+   `dc`, `distclean`, `reset` | Removes all generated files, `PandaSpigot-API`, `PandaSpigot-Server`, and base. |
```

uRyanxD (changes requested):

panda
anzz1: That naming would be confusing for `edit`. "base" refers literally to base, an unedited version of paper before all the pandaspigot patches/edits. Yes sure "base" could also mean the panda base before edits, but it just makes it confusing because then it'd mean two different things. I think it's how it is for a reason, separating the paper and panda, and good as-is.

uRyanxD (approved):
Sounds good to me—thanks!

PR #435, opened 2026-07-09
[WIP] Backport modern Animals API & Extend Animals API

anzz1:
- Backport modern Animals API

add methods:
  UUID getBreedCause();
  void setBreedCause(UUID uuid);
  boolean isLoveMode();
  int getLoveModeTicks();
  void setLoveModeTicks(int ticks);
  boolean isBreedItem(ItemStack stack);
  boolean isBreedItem(Material material);

- Extend Animals API

add new methods:
  Material[] getBreedItems()
  void setBreedItems(Material[] materials)
  void addBreedItem(Material material)
  void removeBreedItem(Material material)
  
  **NOTE:** When merging, use "rebase & merge", not "squash & merge", so the backport of modern animals api and the new pandaspigot extension stay as their separate commits.

### DIFF
--- a/patches/api/0032-Backport-modern-Animals-API.patch
+++ b/patches/api/0032-Backport-modern-Animals-API.patch
@@ -1,0 +1,85 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Mon, 13 Jul 2026 05:55:48 +0000
+Subject: [PATCH] Backport modern Animals API
+
+
+diff --git a/src/main/java/org/bukkit/entity/Animals.java b/src/main/java/org/bukkit/entity/Animals.java
+index f0dc157dc3f228cd9e92b804feb3e3e4396e1e43..46a11d900bec03947dd3351cbf69bf43d4cdd22d 100644
+--- a/src/main/java/org/bukkit/entity/Animals.java
++++ b/src/main/java/org/bukkit/entity/Animals.java
+@@ -1,6 +1,73 @@
+ package org.bukkit.entity;
+ 
++import java.util.UUID;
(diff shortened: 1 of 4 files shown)


anzz1:
Hmmm, I'm stuck in bit of a problem. With the following very simple example plugin:
```java
package com.example.animals;

import org.bukkit.Material;
import org.bukkit.entity.EntityType;
import org.bukkit.plugin.java.JavaPlugin;

public class PigsEatPotatoes extends JavaPlugin {

    @Override
    public void onEnable() {
        getServer().getAnimalData().addBreedItem(EntityType.PIG, Material.POTATO);
    }
}
```

[PigsEatPotatoes.jar.zip](https://github.com/user-attachments/files/30065699/PigsEatPotatoes.jar.zip)

The pigs can now be fed potatoes to breed them, however the PathFinderGoalTempt doesn't work with potatoes, only with carrots, as if nothing was changed. Which is strange because it means the tempt logic itself isn't broken as the carrot still works, yet somehow the added potato doesn't, and I don't see where the problem is. How can the feeding work, yet tempting not?

MasterDash5:
You're checking `this.k == null` in PathfinderGoalTempt. EntityPig sets that to `Items.CARROT` in its constructor. Edit: ~~The field `k` being null means to default to wheat.~~ This is wrong.

anzz1:
Yeah, originally it is `Items.CARROT` but I specifically set it to `null` here:  
https://github.com/hpfxd/PandaSpigot/pull/435/files#diff-009c9a418f3cb98ae809abb9e24ef89f6ecc0c668a3be908881d6bbf4532418cR255-R256

And `null` item means that is should use the `getTemptationItems()` method instead. Yet somehow it still only works with the original carrot. I'm stumped

PR #436, opened 2026-07-10
[WIP] Fix MC-2518: No breaking sound or particle effect for some tools and all armor

uRyanxD:
[X] Fixed hoe-breaking sound
[ ] Fixed the hoe-breaking particle

[X] Fixed sword-breaking sound
[ ] Fixed the sword-breaking particle

[X] Fixed fishing-breaking rod sound
[ ] Fixed the fishing rod-breaking particle

[X] Fixed armor-breaking sound
[ ] Fixed the fishing armor-breaking particle

I need to retest whether these are truly the only affected items, since it’s been over a month since I commented on the issue. Also add comments

### DIFF
--- a/patches/server/0132-Fix-MC-2518-No-breaking-sound-or-particle-effect-for.patch
+++ b/patches/server/0132-Fix-MC-2518-No-breaking-sound-or-particle-effect-for.patch
@@ -6,36 +6,38 @@ Subject: [PATCH] Fix MC-2518: No breaking sound or particle effect for some
 
 
 diff --git a/src/main/java/net/minecraft/server/Entity.java b/src/main/java/net/minecraft/server/Entity.java
-index 9d8bb72b469163db7def78f623d18dc4056b1f6b..d55ad2cb8836ea552774803a0874d112195d4b4e 100644
+index 9d8bb72b469163db7def78f623d18dc4056b1f6b..e5da5d5d8ec2e25d02847a8db1dcb8df3f237080 100644
 --- a/src/main/java/net/minecraft/server/Entity.java
 +++ b/src/main/java/net/minecraft/server/Entity.java
 @@ -858,6 +858,8 @@ public abstract class Entity implements ICommandListener {
  
      }
  
-+    public void playSound(String s, float f, float f1) {}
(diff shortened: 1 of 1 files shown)


anzz1:
i don't think you need to change that much, if what is wanted is to always broadcast the item breaking sound and particles (sometimes causing duplicates, but it doesn't matter), could just extend #412 in the renderbrokenitemstack to always broadcast, i.e. replace "hasUnbreaking" with "true" in the patch

in other words, changing
```diff
-entityliving.renderBrokenItemStack(this, hasUnbreaking); // PandaSpigot - Unbreaking Break Effect Fix
+entityliving.renderBrokenItemStack(this, true); // PandaSpigot - Unbreaking Break Effect Fix
```
should do it

there is also no need for new method with "includesource" simply call the base playsound, i.e. super.playsound instead of this.playsound in entityliving, because the base playsound in entity does already exactly this. its only the playsound in entityhuman that is special which deliberately doesnt play sounds for the source who is meant to play them client side

anzz1:
but like mentioned in the other PR this is a brute force way to do it, it could be possible to figure out what actually causes the desync. for unbreaking the brute force way is the *only* way due to the RNG involved, but for the regular breaking stuff there *could* be a way to fix it without brute forcing the issue and causing the duplicate sounds&particles whenever the client is *not* desynced

uRyanxD:
> but like mentioned in the other PR this is a brute force way to do it, it could be possible to figure out what actually causes the desync. for unbreaking the brute force way is the _only_ way due to the RNG involved, but for the regular breaking stuff there _could_ be a way to fix it without brute forcing the issue and causing the duplicate sounds&particles whenever the client is _not_ desynced

I didn't do that because then `PlayerSoundEvent` wouldn't be called; `EntitySoundEvent` would be instead.

I agree that this is a brute-force method, and there might be a better way to handle this case.

I'm going to close this PR. What do you think about marking your PR as WIP until a fix for this issue is included? If you agree, I'll review your other PRs tomorrow and merge a few of them.

anzz1:
I agree, the [MC-1040](https://bugs.mojang.com/browse/MC/issues/MC-1040) and [MC-2518](https://bugs.mojang.com/browse/MC/issues/MC-2518) are linked issues and should be fixed at same time. Let's investigate further in #412

PR #437, opened 2026-07-11
Ensure chunk packet before login teleport

### LINKED ISSUE #408: World is invisible for a bit on join
On join, the world is invisible for ~1 second while you see the chunks lazily loading in. In the screenshot why its half-visible I assume that its due to the worldspawn chunk always being loaded in and I'm next to it.

This doesn't happen in vanilla and is quite irritating plus you can see inside the ground where you're not supposed to essentially showing whats under you.
It should be possible, for example to delay the login event so from the client's perspective he's still connecting until the chunk he's spawning in is loaded? So you'll never see the world like this in half-loaded state.
This probably should be a configuration option if immediate login and spawning is preferred over making sure the world is loaded in first.

Image

anzz1:
This patch makes sure that the chunk packet for the player position is sent first before the teleport packet to it, this prevents the issue where client thinks it's ready and spawns the player into a void before having received the correct chunk packet

Fixes #408

### DIFF
--- a/patches/server/0141-Ensure-chunk-packet-before-login-teleport.patch
+++ b/patches/server/0141-Ensure-chunk-packet-before-login-teleport.patch
@@ -1,0 +1,146 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Sat, 11 Jul 2026 04:43:23 +0000
+Subject: [PATCH] Ensure chunk packet before login teleport
+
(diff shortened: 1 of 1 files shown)


anzz1:
Of course, this is something that the client should check and not the server, but we can't fix the client so this is the next best thing.

Or actually, its both a server and client problem. Server shouldn't teleport a player to somewhere before it has sent the client the place to teleport to, but also the client shouldn't show itself off the "Downloading terrain ..." until it has actually downloaded the terrain.

Any case, this solves the issue without having to change the client.

PR #438, opened 2026-07-15
ci: upload API artifact

anzz1:
Needs https://github.com/dashnetworkxyz/PandaSpigot/commit/9f28abc181c00b3fb700c2c05dfed9bf92e49f34  and https://github.com/dashnetworkxyz/PandaSpigot/commit/f54e6b0dd46706a3eddbee092593e46a186dd881 merged for this to make sense.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,9 +37,20 @@ jobs:
         run: |
           cp -v "./paperclip/build/libs/paperclip-1.8.8-R0.1-SNAPSHOT.jar" "./pandaspigot-$GITHUB_RUN_NUMBER.jar"
 
-      - name: Upload artifacts
+      - name: Copy API jar
+        run: |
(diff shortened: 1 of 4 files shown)


MasterDash5:
It doesn't make much sense to publish an API jar without any of its libraries included. This jar is really meant for use with maven.

anzz1:
It works perfectly as long as https://github.com/dashnetworkxyz/PandaSpigot/commit/9f28abc181c00b3fb700c2c05dfed9bf92e49f34 and https://github.com/dashnetworkxyz/PandaSpigot/commit/f54e6b0dd46706a3eddbee092593e46a186dd881 is merged.
And yes indeed the API jar needs its libraries included so it works like the original Spigot API jar which is also published, so you are able to make plugins with just the API jar.
I edited the description to reflect this.

PR #439, opened 2026-07-18
Unstuck entities from fences

### LINKED ISSUE #431: Cows get stuck on fence
Cows get stuck on the corner of fenced-in pens inside the fence and can't be pushed out nor do their roaming mechanics work anymore. After a time more and more cows get stuck. Presumably this is due to logging out and chunk unloading and then loading with the cow on the corner and the fence doesn't properly push the intersecting cow out of the fence during load and it gets stuck.

This is probably a vanilla bug but it'd be nice to fix nevertheless.

Image

anzz1:
Fixes #431

### DIFF
--- a/patches/server/0142-Unstuck-entities-from-fences.patch
+++ b/patches/server/0142-Unstuck-entities-from-fences.patch
@@ -1,0 +1,120 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: anzz1 <anzz1@live.com>
+Date: Tue, 21 Jul 2026 01:52:56 +0000
+Subject: [PATCH] Unstuck entities from fences
+
+
+diff --git a/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java b/src/main/java/com/hpfxd/pandaspigot/config/PandaSpigotWorldConfig.java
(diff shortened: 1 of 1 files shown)


MasterDash5:
Fences, fence gates, and cobble walls aren't the only things a mob can get stuck in. They can get stuck in any block that doesn't have a fully horizontal block size. An example would be iron bars and glass panes. I believe they can even get stuck on a dragon egg.

anzz1:
I figured so too and tested with doors, iron gates, and glass panes. However I weren't able to get the cows I tested with stuck in any of them. Probably because doors are on one side of of a block, while I didn't really know why glass panes or iron bars didn't work, maybe because they have a narrower collision box.

It could have been an error on my part though to only test with cows as I later found they were less easy to get stuck, for example the `if (inside)` wasn't necessary at all for cows because they never got so far stuck, but chickens on the other hand were very easy to get inside fence either by pushing or themselves.

There is no reason this wouldn't work for the other block types too, I'll try to get something stuck in a glass pane or iron bar and change the check to include them too.

anzz1:
stuckchikens
Oh yeah, was able to get chikens stuck on BlockPane type so iron bars / glass panes, as long as the above block was blocked so they couldn't just climb out.
So those should be def added, with additional caveat that only care about them if the block above is a non-traversable block.

anzz1:
Another way to do this I guess could be to set all the non-full cube collision boxes to be full cubes as far as NPCs are concerned, although idk which is the better way. This one can be done fairly simply at least.

PR #440, opened 2026-07-30
Fix bottles moving slots with cauldrons

MasterDash5:
The current way emptying a cauldron with a single glass bottle will result in that bottle either moving slots, or being thrown out of your inventory if no slots are open. Mojang decided to give you the water bottle first, then remove the old glass bottle.

This PR adjusts this logic to directly set the glass bottle to a water bottle if there is only 1 glass bottle. This matches what modern versions do, as well as how removing banner patterns is done.


anzz1:
Nice, you're upstreaming fixes from DashSpigot to here, fantastic :+1:

Imo almost everything from DashSpigot could be upstreamed, I do get that its bit of work to do that tho :smile: 
Though maybe the patch that removes BlockPhysicsEvent should just remove it for redstone like in 1.9.4 version of spigot: https://github.com/PaperMC/Paper-archive/blob/ver/1.9.4/Spigot-Server-Patches/0106-Option-to-disable-BlockPhysicsEvent-for-Redstone.patch

PR #441, opened 2026-07-30
Use valueOf for Entity DataWatchers

MasterDash5:
This PR adjusts entity's to consistently use `valueOf` for initialized data in the `h()` method. Most entities already use valueOf, this just cleans up the rest.

This saves a tiny bit of memory, and also removes deprecation warnings during compilation.

### DIFF
--- a/patches/server/0134-use-valueOf-in-DataWatcher.patch
+++ b/patches/server/0134-use-valueOf-in-DataWatcher.patch
@@ -1,0 +1,185 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: MasterDash5 <github@masterdash5.com>
+Date: Thu, 30 Jul 2026 17:35:17 -0600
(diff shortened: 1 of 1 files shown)


uRyanxD:
Is `valueOf` really necessary?

Creeper, for example, doesn't use it.
https://prnt.sc/4g6AGvdfAGRW

MasterDash5:
I don't see that in my codebase, they all use valueOf for me. They do need be wrapped so DataWatcher can lookup the class type.

uRyanxD:
> I don't see that in my codebase, they all use valueOf for me. They do need be wrapped so DataWatcher can lookup the class type.

Oh right, I was looking at the decompiled version using an updated Fernflower, but this class uses the version inherited from Spigot.

IntelliJ doesn't seem to flag this as a syntax error; I believe Java automatically adds `valueOf` during compilation. Either way, it's fine as is.

uRyanxD (approved):
LGTM

PR #442, opened 2026-07-31
Fix MC-12211

MasterDash5:
This PR fixes [MC-12211](https://bugs.mojang.com/browse/MC/issues/MC-12211)

The method `e(world, blockposition, iblockdata)` only works for compare mode. This just adds the correct logic for subtract mode into the flag.


uRyanxD:
That works, but wouldn't that cause the block to schedule an unnecessary tick? Because the e() method wasn't fixed, so g() will keep using the buggy version

PR #443, opened 2026-08-04
Fix MC-7809 Slabs, stairs, and trapdoors placing incorrectly

ThatApplePieGuy:
Fixes [MC-7809](https://bugs.mojang.com/browse/MC/issues/MC-7809), which causes slabs, stairs, and trapdoors to get placed with the wrong orientation. This is most noticeable when bridging with slabs
https://imgur.com/bridging-with-top-slabs-smp-P5vhAOI

The server determines the orientation of these blocks based on the cursor y position of the block placement packet. The issue is the client can place against the top half of a block but sends a cursor y of 0.5 after rounding, which the server treats as the bottom half

[And for completeness a plugin version of the fix](https://github.com/ThatApplePieGuy/PacketPatcher/blob/main/src/main/java/com/github/thatapplepieguy/packetpatcher/fix/BlockOrientationFix.java)


anzz1:
Looks good!

There is just one little thing I would change, use the full name in comment
```diff
-// PandaSpigot - Fix MC-7809
+// PandaSpigot - Fix MC-7809 Slabs, stairs, and trapdoors placing incorrectly
```
So people perusing the code know what was fixed at glance without having to look up the issue code

uRyanxD (approved):
LGTM, just needs to be rebased

PR #444, opened 2026-08-08
Several small quality-of-life improvements

uRyanxD:
CI: The "Apply patches" step has been removed, as the "Setup" step already applies them. The .jar file is no longer compressed within a .rar file.
README: Minor simplifications.
Conventions: Added a missing Gradle flag and moved some items from there to PandaSpigot-Server, where they are actually relevant.
Other: Updated gradlew; `./panda jar` now only compiles the project and generates the paperclip jar (as intended).
Adventure: Updated to the last version prior to the 5.x series, which no longer supports older versions of Minecraft.


anzz1:
Are gradle and the shadow plugin `com.gradleup.shadow` versions linked, or is it just a coincidence?
As I noticed that previously gradle was version 9.4.1 and shadow version was 9.4.1, and now gradle is version 9.7.0 and shadow is version 9.6.1

uRyanxD:
> Are gradle and the shadow plugin `com.gradleup.shadow` versions linked, or is it just a coincidence?
> As I noticed that previously gradle was version 9.4.1 and shadow version was 9.4.1, and now gradle is version 9.7.0 and shadow is version 9.6.1

Not necessarily; I usually update because there have been times when Gradle broke or changed something important that affected Shadow. However, that usually only happens with major versions.

Example:
https://prnt.sc/CSIK5VO-4b02

anzz1:
Yeah when you look at the changelog it seems that shadow attempts to somewhat keep in line with the gradle version in its own versioning, but its not exactly 1:1

Gradle changelog: https://gradle.org/releases/
Shadow changelog: https://github.com/GradleUp/shadow/blob/main/docs/changes/README.md

And even so, shadow seems to generally keep backwards compatibility with few versions down: https://github.com/GradleUp/shadow#compatibility-matrix

It seems that in this case, Gradle 9.7.0 is so new (released 4 days ago) that a matching shadow 9.7.0 is still in development and 9.6.1 is the latest stable version.

uRyanxD:
> Yeah when you look at the changelog it seems that shadow attempts to somewhat keep in line with the gradle version in its own versioning, but its not exactly 1:1
> 
> Gradle changelog: https://gradle.org/releases/ Shadow changelog: https://github.com/GradleUp/shadow/blob/main/docs/changes/README.md
> 
> And even so, shadow seems to generally keep backwards compatibility with few versions down: https://github.com/GradleUp/shadow#compatibility-matrix
> 
> It seems that in this case, Gradle 9.7.0 is so new (released 4 days ago) that a matching shadow 9.7.0 is still in development and 9.6.1 is the latest stable version.

Yes, but there shouldn't be any issues; this version of Shadow had already been updated for Gradle Release Candidate 1.

PR #445, opened 2026-08-10
Backport World.isDayTime()

anzz1:
Backport [World.isDayTime()](https://jd.papermc.io/paper/26.1.2/org/bukkit/World.html#isDayTime())

Also, the patch serves as a consolidation of API backports from Paper small enough to not necessitate their own patch that can be edited and added to in the future, to keep the number of patches manageable


uRyanxD (approved):
LGTM, thank you

PR #446, opened 2026-08-24
Backport Modern Tick Loop

uRyanxD:
So, we originally used the Modern Tick Loop about 2–3 years ago, but its backport to 1.8 was a mess; it included unnecessary changes and was missing features, among other issues like the vanilla bug https://bugs.mojang.com/browse/MC/issues/MC-183518, etc.

That’s why I had temporarily reverted to the current tick loop. Now, I’m re-adding the Modern Tick Loop—this time fully functional and ported by me.

### DIFF
--- a/patches/server/0104-Backport-modern-tick-loop-system.patch
+++ b/patches/server/0104-Backport-modern-tick-loop-system.patch
@@ -3,22 +3,225 @@ From: uRyanxD <familiarodrigues123ro@gmail.com>
 Date: Mon, 10 Feb 2025 14:55:55 -0300
 Subject: [PATCH] Backport modern tick loop system
 
-Also removes redundant empty list check before putting a task on the
-main thread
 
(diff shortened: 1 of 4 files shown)


uRyanxD:
If anyone wants to jump ahead and test this PR before it gets merged (there are a few PRs in the queue ahead of this one), here is a download: https://www.mediafire.com/file/0j29n5kjc5y4kn6/paperclip.jar/file

kalibrieren:
`// Pan`? xd

uRyanxD:
> `// Pan`? xd

lol

There's nothing I can do xD; Git controls that.

kalibrieren:
you probably shouldn't comment out `postToMainThread`/`isMainThread` directly in `WorldServer`, just remove the overrides like we did in WindSpigot

uRyanxD:
> you probably shouldn't comment out `postToMainThread`/`isMainThread` directly in `WorldServer`, just remove the overrides like we did in WindSpigot

Nothing in PandaSpigot uses them, so for us, it can stay as is—no problem

PR #447, opened 2026-09-03
Fix title times in Adventure

MasterDash5:
This PR fixes title times in the built-in Adventure API. 1.8 expects the times in ticks, however we had this in nanoseconds instead.


uRyanxD (approved):
LGTM

PR #448, opened 2026-09-04
Drop the "Configurable time update frequency" patch

uRyanxD:
Fixes an issue that prevented some client-side mods from correctly calculating the server TPS.

The patch changed the default game-time update frequency, thereby breaking calculations used by some mods; making this configurable doesn't make much sense either, as anzz1 mentioned in another PR.

---

I also added the native Adventure API as a highlight in the README.


anzz1:
I fully agree, 1 small packet per second per client to update the time is ultimately nothing.

